PR #33, opened 2024-10-22
Readable file around `object_store::BufReader`

kylebarron:
"File-like object" wrapper


kylebarron:
cc @martindurant this adds a file-like object wrapper, which we could also wrap in fsspec

object-store-rs/src/buffered.rs
```diff
@@ -0,0 +1,212 @@
...
+    let store = store.into_inner();
+    let runtime = get_runtime(py)?;
+    let meta = py.allow_threads(|| runtime.block_on(store.head(&path.into())))?;
+    Ok(PyReadableFile(Arc::new(Mutex::new(BufReader::new(
+        store, &meta,
+    )))))
+}
+
+#[pyfunction]
+pub(crate) fn open_async(py: Python, store: PyObjectStore, path: String) -> PyResult<Bound<PyAny>> {
```
martindurant, line 30: I'm not sure you want to bother with open_async for now - if it's to be a file-like API, may as well just do that, and only give the blocking methods. It is a stateful object, after all.

kylebarron: Presently, this PR only provides async methods, not blocking methods. The API is meant to be similar to https://github.com/Tinche/aiofiles.

It would be a good enhancement to have two separate Python classes, one returned by `open` and another returned by `open_async`. Then the methods on the object returned by `open_async` would be named `read()` (without the `async` suffix) but still return coroutines.

martindurant: OK, so not really like https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L911 (implementations elsewhere)

kylebarron: No, sorry, this wasn't modeled _after_ fsspec, but it could be used in the fsspec wrapper.

PR #40, opened 2024-10-23
Support range in GetOptions

kylebarron:
Closes https://github.com/developmentseed/object-store-rs/issues/10

### DIFF
--- a/object-store-rs/python/object_store_rs/_get.pyi
+++ b/object-store-rs/python/object_store_rs/_get.pyi
@@ -1,6 +1,6 @@
 import sys
 from datetime import datetime
-from typing import List, Sequence, TypedDict
+from typing import List, Sequence, Tuple, TypedDict
 
 from ._list import ObjectMeta
 from .store import ObjectStore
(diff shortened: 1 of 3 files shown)


kylebarron:
From https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2433392608:

> Note that they can can be negative, so a suffix range would be `(-N, None)`.

@martindurant for now this PR uses `(None, N)` for a suffix request, where `N` is positive. This is documented in the docstring.

We may change this in the future to be a dict, with keys like `offset`, `bounded`, and `suffix`, to more closely match the upstream type: https://docs.rs/object_store/latest/object_store/enum.GetRange.html

martindurant:
> for now this PR uses (None, N)

That looks a lot like "first N bytes" to me, not suffix :|  I could be biased.

kylebarron:
I think in general it's too easy to get confused by these. But there's no reason to use `(None, N)` when you can easily by explicit about `(0, N)`.

PR #49, opened 2024-10-25
split put and put_multipart

kylebarron:


kylebarron:
Instead of doing this, I'm going to try and keep them the same function. The only difference is that [`PutMode`](https://docs.rs/object_store/latest/object_store/enum.PutMode.html) only exists for `put` and not `put_multipart`

Issue #47, opened 2024-10-24
Streaming uploads


kylebarron:
We should allow async iterator input into `put`. Then we can pass the result of `get` directly into `put` and have streaming transfers from one store to another.

machichima:
This feature would be extremely helpful! We're really looking forward to it. Is there any update on this issue?

kylebarron:
I attempted it in https://github.com/developmentseed/obstore/pull/54. It compiles fine, but then appeared to hang in my manual tests. I'm not sure what the issue is and haven't had time to look back into it. If you're interested, you can go to that branch and test it out (you can just `pip install .` from the obstore directory as long as you have Rust installed)

kylebarron:
It turned out that the only reason for my errors in #54 was that I was hitting my request timeout. Raising the timeout shows that the PR does in fact work.

rupurt:
You absolute legend @kylebarron! If we ever meet in person I owe you a 6 pack 🍻

PR #54, opened 2024-10-25
Async stream upload

kylebarron:
### Change list

- Add support for sync iterators and async iterators in `obstore.put`.
    - Improved `PutInput` concept, which is an enum over pull-based sources, sync push-based sources, and async push-based sources.
    - Previous accepted input into `PutInput` is now contained in `PullSource`, including files on disk, file-like objects, and buffers.
    - Now there's both `SyncPushSource`, which covers sync iterators emitting buffer protocol objects, and `AsyncPushSource`, which covers async iterators emitting buffer protocol objects.
- Add tests.
- More detailed exceptions by using `Debug` instead of `Display` on the rust side.
- Add note in `BytesStream` docstring about making the timeout parameter longer.
- Rename `Buffer` to `Bytes` in type hints (from https://github.com/developmentseed/obstore/pull/128)
- Add `BytesStream` to website docs
- Update documentation

Closes https://github.com/developmentseed/obstore/issues/47


kylebarron:
It turns out that the error I was previously hitting was just the request timeout, since I was downloading a large file.

PR #60, opened 2024-10-30
Kyle/fsspec integration

martindurant:
Running this test fails: 
```
Exception: Generic S3 error: Error performing list request: Error after 0 retries in 39.083µs, max_retries:10, retry_timeout:180s, source:builder error for url (http://localhost:5555/test?delimiter=%2F&list-type=2)
```
even though that URI is gettable with requests, returning valid XML. Thoughts?


kylebarron:
> even though that URI is gettable with requests, returning valid XML. Thoughts?

I think that usually means that it's falling back to instance authentication? Maybe we should default that to off for simpler errors

martindurant:
> we should default that to off

How to do that?

kylebarron:
I think there'll be a bit of a learning curve for us to figure out the right set of auth configuration and ensure it's documented correctly. 

So far, whenever I've hit auth errors, I've been able to re-check my credentials and get it working

object-store-rs/python/object_store_rs/fsspec.py
```diff
@@ -26,16 +26,12 @@
...
-    from object_store_rs.store import ObjectStore
-
 
 class AsyncFsspecStore(fsspec.asyn.AsyncFileSystem):
-    store: ObjectStore
 
     def __init__(
         self,
-        store: ObjectStore,
+        store,
```
kylebarron, line 34: Why did you remove the `ObjectStore` typing here?

kylebarron:
https://github.com/developmentseed/obstore/pull/62 has an example of working tests against `moto`

martindurant:
Perfect, I'll build on that. "Unsigned" seems wrong, but oh well.

object-store-rs/python/object_store_rs/fsspec.py
kylebarron: Yeah it's a type-only object. Perhaps it should be changed in the future to be an actual Python object that can be imported at runtime. The `from __future__ import annotations` should allow this to not be evaluated.

martindurant:
Er, something has gone *very* wrong with my git here. I'll get back to it tomorrow with any luck.

kylebarron:
The branch you're trying to merge into is also behind. It's fine if you switch the target branch to `main`

kylebarron:
I'll close this as I believe the only PR you intend to work on is #63, but reopen if I'm wrong.

PR #63, opened 2024-10-31
fsspec integration

martindurant:
Replaces #60


kylebarron:
I'm afraid I don't know enough about pytest to understand why the test setup isn't working correctly. Does it work if you move it into `conftest`?

tests/test_fsspec.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+pytest.importorskip("moto")
```
kylebarron, line 3: I'm happy to make `moto` a dev-dependency, so you can take this line out.

martindurant: I already is, but I didn't realise that :)

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,159 @@
...
+> The ObjectStore interface is designed to mirror the APIs of object stores and not filesystems, and thus has stateless APIs instead of cursor based interfaces such as Read or Seek available in filesystems.
+>
+> This design provides the following advantages:
+>
+> - All operations are atomic, and readers cannot observe partial and/or failed writes
+> - Methods map directly to object store APIs, providing both efficiency and predictability
+> - Abstracts away filesystem and operating system specific quirks, ensuring portability
+> - Allows for functionality not native to filesystems, such as operation preconditions and atomic multipart uploads
+
+Where possible, implementations should use the underlying `object-store-rs` APIs
```
kylebarron, line 14:
```suggestion
Where possible, implementations should use the underlying `obstore` APIs
```

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,159 @@
...
+
+    async def _info(self, path, **kwargs):
+        head = await obs.head_async(self.store, path)
+        return {
+            # Required of `info`: (?)
+            "name": head["path"],
+            "size": head["size"],
+            "type": "directory" if head["path"].endswith("/") else "file",
+            # Implementation-specific keys
+            "e_tag": head["e_tag"],
```
kylebarron, line 120: We have two different capitalizations of `e_tag` and `ETag`. Is one better than the other?

martindurant: It depends on how closely you want to match hat fsspec gives you. It had ETag, which is what AWS calls it in the response data. I don't mind...

kylebarron: This integration should match fsspec semantics as closely as possible. The rest of obstore will probably stick with returning `e_tag` because that's what `object_store` uses under the hood. But if fsspec users would expect `ETag`, then we should change it here.

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,159 @@
...
+            return await resp.bytes_async()
+
+        if start is not None and end is not None:
+            return await obs.get_range_async(
+                self.store, path, offset=start, length=end - start
+            )
+
+        raise NotImplementedError("todo: handle open-ended ranges")
+
+    async def _cat_ranges(
```
kylebarron, line 67: Would you be able to add some tests for this? I think this function is the least stable of everything because it relies on custom code to split the ranges into per-file ranges and piece the outputs together again.

We should have some tests for multiple ranges in a single file, single range for multiple files, multiple ranges for multiple files, and maybe overlapping ranges too.

martindurant: I'll look into this. I strongly suspect that it will "just work" with the upstream fsspec code, actually, rather than having to carve out the per-file requests and repeat the code in rust. The only downside of that would be bytes copies in python.

kylebarron: > I strongly suspect that it will "just work" with the upstream fsspec code, actually,

Yes, it would just work with the upstream code, but the benefit here is that we get request merging in the underlying Rust code. We should absolutely use the underlying `get_ranges` if we can; we just have to map between multi-file and single-file ranges.

martindurant: > request merging in the underlying Rust code

_as_opposed_ to doing it in python, you mean? I wonder if it makes a practical difference.

kylebarron: True. My position is mostly: we should use whatever tools `object_store` provides us, because the Rust code I'd guess will be _slightly_ faster, which could make a difference for large-scale data loading with `get_ranges`, like in zarr.

I don't think it's too much work to have this helper function, and it would also be useful in places that don't depend on `fsspec`, like in the zarr PR: https://github.com/zarr-developers/zarr-python/pull/1661/files#diff-6f6bc4f5bf8c4e9eb8f9bb486d3699e17e4a4d6efc90b922c63a75d45cd7a9a6R47-R51

martindurant: OK, I can agree with that reasoning.

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,159 @@
...
+import fsspec.asyn
+import fsspec.spec
+
+import obstore as obs
+
+
+class AsyncFsspecStore(fsspec.asyn.AsyncFileSystem):
+    def __init__(
+        self,
+        store,
```
kylebarron, line 34: Before this is merged I'd like to restore the typing here with `store: ObjectStore`

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,159 @@
...
+        *args,
+        asynchronous=False,
+        loop=None,
```
kylebarron, line 37: What's the right type for `loop`?

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,159 @@
...
+from collections import defaultdict
+from typing import Any, Coroutine, Dict, List, Tuple
+
+import fsspec.asyn
+import fsspec.spec
+
+import obstore as obs
+
+
+class AsyncFsspecStore(fsspec.asyn.AsyncFileSystem):
```
kylebarron, line 31: We should add docstrings for any public methods. I suppose that's just the class-level docstring + `__init__`.

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,159 @@
...
+        if start is None and end is None:
+            resp = await obs.get_async(self.store, path)
+            return await resp.bytes_async()
+
+        if start is not None and end is not None:
+            return await obs.get_range_async(
+                self.store, path, offset=start, length=end - start
+            )
+
+        raise NotImplementedError("todo: handle open-ended ranges")
```
kylebarron, line 65: This was re-implemented in https://github.com/developmentseed/obstore/pull/67, with clearer semantics for offset and suffix requests. Would you be able to map the fsspec range semantics onto this?

martindurant: I just pushed what I have at the moment, so can do whatever you like. 

Note that in the current state of the code, ranges_one and ranges_two both pass separately, but not when run in the same session ("Generic S3 error" on the upload), and I don't know why.

martindurant:
Actually, all the tests are failing in CI, not what I see locally :|

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,185 @@
...
+
+    async def _cat_file(self, path, start=None, end=None, **kwargs):
+        if start is None and end is None:
+            resp = await obs.get_async(self.store, path)
+            return await resp.bytes_async()
+
+        if start is not None and end is not None:
+            return await obs.get_range_async(
+                self.store, path, offset=start, length=end - start
+            )
```
kylebarron, line 80: Also heads up/note to self that this needs to be updated for #71

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,185 @@
...
+        for idx, (path, start, end) in enumerate(zip(paths, starts, ends)):
+            per_file_requests[path].append((start, end, idx))
+
+        futs: List[Coroutine[Any, Any, List[bytes]]] = []
+        for path, ranges in per_file_requests.items():
+            offsets = [r[0] for r in ranges]
+            lengths = [r[1] - r[0] for r in ranges]
+            fut = obs.get_ranges_async(
+                self.store, path, offsets=offsets, lengths=lengths
+            )
```
kylebarron, line 107: Ditto heads up/note to self that this needs to be updated for #71

obstore/python/obstore/fsspec.py
kylebarron: Right now in the code on main there's no clearing of the moto state between test fixtures. I had run into some issue with that. So that could be a place to start.

martindurant:
green :)

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,181 @@
...
+        result = await obs.list_with_delimiter_async(self.store, path)
+        objects = result["objects"]
+        prefs = result["common_prefixes"]
+        if detail:
+            return [
+                {
+                    "name": object["path"],
+                    "size": object["size"],
+                    "type": "file",
+                    "ETag": object["e_tag"],
```
kylebarron, line 150: This should be `e_tag` for consistency

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,181 @@
...
+                }
+                for object in objects
+            ] + [{"name": object, "size": 0, "type": "directory"} for object in prefs]
+        else:
+            return sorted([object["path"] for object in objects] + prefs)
+
+    def _open(self, path, mode="rb", **kwargs):
+        """Return raw bytes-mode file-like from the file-system"""
+        out = BufferedFileSimple(self, path, mode)
+        return out
```
kylebarron, line 160: Is this what's going to have a default implementation when https://github.com/fsspec/filesystem_spec/pull/1732 is merged? Or this is separate?

tests/test_fsspec.py
```diff
@@ -0,0 +1,103 @@
...
+    out = fs.cat_ranges(["data1", "data1"], [10, 30], [20, 60])
+    assert out == [data1[10:20], data1[30:60]]
+
+    # two overlapping ranges, one file
+    out = fs.cat_ranges(["data1", "data1"], [10, 15], [20, 60])
+    assert out == [data1[10:20], data1[15:60]]
+
+    # completely overlapping ranges, one file
+    out = fs.cat_ranges(["data1", "data1"], [10, 0], [20, 60])
+    assert out == [data1[10:20], data1[0:60]]
```
kylebarron, line 88: Are offset and suffix requests valid in `cat_ranges`? If so, we should add tests with those here too. E.g. 

```py
    out = fs.cat_ranges(["data1", "data1"], [10, None], [-20, None])
```

obstore/python/obstore/fsspec.py
martindurant: I thought we discussed this and agreed ETag?

tests/test_fsspec.py
martindurant: The same conditions would apply as for cat_file, which for this implementation ATM is either None/None or int/int and no negatives. Otherwise we would need an extra step to first find the size of each file, and/or explicitly make suffix Range headers. I think this is OK as is for a start.

obstore/python/obstore/fsspec.py
martindurant: It could have been done with that PR, once release. 

I was worried about passing Buffers around rather than bytes and/or whether read buffering might be implemented on the rust side (without copies). I can confirm that AbstractBufferedFile works directly including the fsspec change. 

However, we *will* end up making a class to deal with writing data unless we only every write whole memory buffers in a single go as opposed to multi-part.

obstore/python/obstore/fsspec.py
kylebarron: The rest of the codebase already uses `e_tag` (including elsewhere in this PR)

tests/test_fsspec.py
kylebarron: > explicitly make suffix Range headers

Yeah that's what I had in mind.

obstore/python/obstore/fsspec.py
kylebarron: > However, we _will_ end up making a class to deal with writing data unless we only every write whole memory buffers in a single go as opposed to multi-part.

We can wrap a `BufWriter` similarly to what we do now with `BufReader`, which will use a multi part upload under the hood.

obstore/python/obstore/fsspec.py
kylebarron: > if fsspec users would expect `ETag`, then we should change it here.

It looks like [s3fs uses `ETag`](https://github.com/fsspec/s3fs/blob/f3f63cbfbfe71a4355abd63cafd8c678c4a5a0af/s3fs/tests/test_s3fs.py#L2468), but [adlfs uses `etag`](https://github.com/fsspec/adlfs/blob/7f06dbdd410224047df09d0ef9e5e9913c64bf7d/adlfs/spec.py#L61). I would've hoped for more consistency here. Ideally this should be defined in the fsspec spec.

martindurant: We could take this chance to make a guarantee about available fields... fsspec does have the `ukey` method on a filesystem that can be the etag output, but it would be best to have some guarantees about field names and types from here on. That would probably mean making dict-like subclasses for each (most) implementations, which derives something like etag and created time from whatever the remote gives.

martindurant: (GCS provides "etag", as well as md5 and crc)

kylebarron: In the absence of a standard from fsspec, I'd opt to use `e_tag` to be consistent with the rest of the library, and we can make a breaking change later. `ETag` is still used here: https://github.com/developmentseed/obstore/pull/63/files/0a0a2fc7d7119c290b33e1b1985f625dfbb2b79e#diff-36786859c2825a3983ff9208d66ac33cc191b38bf6fc1e562a56a2b073ae229cR135

martindurant:
(this fails because he _fetch_range fix in fsspec isn't released yet)

martindurant:
OK, so the prefix case doesn't work ATM (I wasn't sure if you were implying it should) and "create" doesn't raise for moto. I haven't tried real s3 yet.

kylebarron:
I don't think `"create"` will work until https://github.com/apache/arrow-rs/pull/6682

martindurant:
OK, I can make that one xfail, and I suppose we wait for fsspec's release before merging here?

kylebarron:
If this doesn't work until the next fsspec release, we should wait to merge until that is released and then probably do an fsspec version check on module import.

martindurant:
It would work now with the subclass I had before, which might prove convenient in the long run... up to you.

kylebarron:
I'm ok with the subclass as long as it's not exposed in a public API. It's essentially just an implementation detail for now, right?

martindurant:
Let's keep to this fromalism for now, then.

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,177 @@
+"""Fsspec integration.
+
+The underlying `object_store` Rust crate cautions against relying too strongly on stateful filesystem representations of object stores:
```
kylebarron, line 3:
```suggestion
The underlying `object_store` Rust crate [cautions](https://docs.rs/object_store/latest/object_store/#why-not-a-filesystem-interface) against relying too strongly on stateful filesystem representations of object stores:
```

obstore/python/obstore/fsspec.py
kylebarron: >        loop: since both fsspec/python and tokio/rust may be using loops, this should
            be kept None for now

If `None` is the only valid argument to `loop`, then we should remove it from `__init__` and manually pass `loop=None` to `super().__init__` below.

kylebarron (commented):
Just a couple more comments, but overall looks good

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,177 @@
...
+            resp = await obs.get_async(self.store, path)
+            return await resp.bytes_async()
+
+        return await obs.get_range_async(self.store, path, start=start, end=end)
+
+    async def _cat_ranges(
+        self,
+        paths: List[str],
+        starts: List[int],
+        ends: List[int],
```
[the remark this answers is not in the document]
martindurant: > Can any element of starts or ends be None?

We can choose to support that. For kerchunk use, we will only have full files (called via cat()) and all known ranges (called via cat_ranges()). Having all that and partial/suffix ranges in one place would be convenient, but I don't think urgent.

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,179 @@
...
+    """An fsspec implementation based on a obstore Store"""
+
+    cachable = False
+
+    def __init__(
+        self,
+        store: obs.store.ObjectStore,
+        *args,
+        asynchronous: bool = False,
+        loop=None,
```
kylebarron, line 41: Remove `loop` too?

obstore/python/obstore/fsspec.py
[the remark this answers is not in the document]
kylebarron: Ok, we can handle that in a follow up.

obstore/python/obstore/fsspec.py
```diff
@@ -0,0 +1,179 @@
...
+        batch_size: int | None = None,
+    ):
+        """Construct a new AsyncFsspecStore
+
+        store: a configured instance of one of the store classes in objstore.store
+        asynchronous: id this instance meant to be be called using the async API? This
+            should only be set to true when running within a coroutine
+        loop: since both fsspec/python and tokio/rust may be using loops, this should
+            be kept None for now
```
kylebarron, line 50:
```suggestion
        loop: since both fsspec/python and tokio/rust may be using loops, this should
            be kept `None` for now, and will not be used.
```

kylebarron:
Publishing v0.3.0-beta.6

PR #64, opened 2024-11-01
Replace medium link by linkedin in docs config

willemarcel:


kylebarron:
Perhaps we should do this in Lonboard too

PR #81, opened 2024-11-08
WIP: requester pays

kylebarron:
This is failing with
```
GenericError: Generic S3 error: Error performing list request: Client error with status 403 Forbidden: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>There were headers present in the request which were not signed</Message><HeadersNotSigned>x-amz-request-payer</HeadersNotSigned><RequestId>QEA9RD5BETE6BN7E</RequestId><HostId>tfgEYvVBmWEx5EZljTRXuR6S6ocChjas5qF8eXThc0yt6bRyGEJCYjCFJ+/YOvA2Q2DT/NSqBTw=</HostId></Error>
```

We should open an issue in `object_store`

Closes #76

PR #94, opened 2024-11-21
Fix retry dict conversion

### LINKED ISSUE #92: How to configure backoff retry
This package is great to have in the python ecosystem, thank you!

I'm having trouble configuring backoff retries. When I try to do something like the following:

```py
from datetime import timedelta
from obstore.store import HTTPStore

store = HTTPStore.from_url(
    "https://...",
    client_options={
        "connect_timeout": "4",
        "timeout": "16",
    },
    retry_config={
        "max_retries": 10,
        "backoff": {
            "base": 2,
            "init_backoff": timedelta(seconds=1),
        },
    },
)
```
I get 
```
    store = HTTPStore.from_url(
            ^^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'backoff'
```

I also attempted to instantiate the config classes like `RetryConfig(max_retries=...)` rather than as dict literals, but when I run `from obstore.store import RetryConfig, BackoffConfig` and variations on that I get an ImportError.

Am I doing something wrong?

kylebarron:
Closes #92 

It's a bit confusing that we have to write timedelta in a string for the `client_options`. https://github.com/developmentseed/obstore/pull/81 may help fix that by adding a `ClientOptions` Python class.

@aldenks You can look at the new test case I added here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1194,7 +1194,7 @@ dependencies = [
 
 [[package]]
 name = "obstore"
-version = "0.3.0-beta.6"
+version = "0.3.0-beta.7"
 dependencies = [
  "arrow",
  "bytes",
--- a/obstore/Cargo.toml
+++ b/obstore/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "obstore"
-version = "0.3.0-beta.6"
+version = "0.3.0-beta.7"
 authors = { workspace = true }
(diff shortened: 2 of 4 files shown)


kylebarron:
I'm pushing 0.3.0-beta.7 if you want to try it out

aldenks:
Tested and working great, thank you Kyle!

PR #115, opened 2025-01-01
add missing py.typed file

gruebel:
## What I am changing

- added missing `py.typed` file to mark the package as typed for type checkers

## How you can test it

- build wheel, install and check it is in the `obstore` folder


kylebarron:
Thank you!

PR #116, opened 2025-01-01
update pre-commit hooks and GHA

gruebel:
## What I am changing

- updated pre-commit hooks versions
- updated GHA versions
- updated `uv` to 0.5.x
- removed `black` from dev deps, because it is not used
- reduced permissions in GH workflows (the other two already had them set)

## How you can test it

- Setup the project via uv 0.5.13

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,13 +17,15 @@ dev-dependencies = [
     "mike>=2.1.3",
     "mkdocs-material[imaging]>=9.5.40",
     "mkdocs>=1.6.1",
-    "mkdocstrings[python]>=0.26.1",
+    "mkdocstrings>=0.27.0",
+    "mkdocstrings-python>=1.13.0",
     "moto[s3,server]>=5.0.18",
     "pandas>=2.2.3",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -9,7 +9,6 @@ dependencies = []
 [tool.uv]
 dev-dependencies = [
     "arro3-core>=0.4.2",
-    "black>=24.10.0",
```
kylebarron, line 12: This is used in the docs to format code docstrings https://mkdocstrings.github.io/python/usage/#options-summary. It looks like the most recent version of mkdocstrings added the option for that to be done using ruff as well as black, so we can remove black as long as we ensure we're using the latest mkdocstrings.

gruebel: Gotcha, will double check

gruebel: @kylebarron done. I lower bound `mkdocstrings` and added `ruff` as dev dependency instead. now both the docs and the source code are using `ruff` as formatter

kylebarron (approved):
Thanks!

Issue #105, opened 2024-12-05
Request metrics


kylebarron:
We should have a middleware that wraps an `Arc<dyn ObjectStore>` and logs request metrics (e.g. # of get, etc)

kylebarron:
Pending #117 it looks like this should be really easy to do.

kylebarron:
Well, it's not! But #117 should still be easy!

kylebarron:
Relevant code in influxdb: https://github.com/influxdata/influxdb3_core/blob/main/object_store_metrics/src/lib.rs

kylebarron:
Ref https://github.com/apache/arrow-rs/issues/7251

PR #122, opened 2025-01-07
wip: tracing/metrics

kylebarron:
For #105 

Well, this is way more complex than I originally thought.

I'm interested in having basic request metrics for an object store: number of requests, their type (get/list/put/delete), timing, and number of bytes downloaded.

At first I was thinking of doing this with a simple `Mutex<Vec<RequestStatistics>>`, but realized it would lead to horrible lock contention when each request wants to append to the vec.

So I've been looking at the `tracing` crate and its [`Instrument`](https://docs.rs/tracing/latest/tracing/trait.Instrument.html), which looks like it might perform better for async code.

However it looks like that's considerably more complex than I want to get into right now. 

In theory, a mature implementation of this would allow the _Python user_ to choose where the traces actually go. Like you could set this up to use ObjectStore but send tracing information to, say, CloudWatch

I'll put this stub up just to record my thoughts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,9 +57,9 @@ dependencies = [
 
 [[package]]
 name = "arrow"
-version = "53.1.0"
+version = "53.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


kylebarron:
Closing in favor of looking at the influxdb implementation: https://github.com/influxdata/influxdb3_core/blob/main/object_store_metrics/src/lib.rs

PR #140, opened 2025-01-14
Implement read_all for async iterable

kylebarron:
### Change list

- Implement `read_all` for `AsyncPushSource`
- Change `read_all` to not copy buffer protocol input.


kylebarron:
I'm not sure why but the tests just hung on [`c7f154f` (#140)](https://github.com/developmentseed/obstore/pull/140/commits/c7f154f9bad35edb390a5c0327f588c9f79a7929). I reverted to using `&mut self` in `read_all` and we'll see if that does better.

Issue #155, opened 2025-01-16
Change `get_range` `start` and `end` parameters to be kwarg only, add `length` as alternative


kylebarron:
Many libraries use either start/end or offset/length for range requests. We should require these params to be keyword-only params.

Additionally, we should allow `length` as a named parameter (exclusive with `end`), for ease of use

This would be a breaking change. Sadly I just released 0.3, so this would require an 0.4.

maxrjones:
FWIW I did not find that there is a significant performance cost to using dataclasses for a range requests in addition to suffix and offset requests in https://github.com/maxrjones/zarr-byterangerequest-microbenchmarks/blob/main/plot-results.ipynb, as long as you use if; then syntax rather than match; case. That was the motivation for moving to only using dataclasses in https://github.com/zarr-developers/zarr-python/pull/2585. Norman suggested [this possible design](https://github.com/zarr-developers/zarr-python/pull/2585#issuecomment-2578172415) for supporting length as well. It would be simple to run similar tests specifically for `obstore` if helpful.

But, requiring kwargs would similarly avoid confusion.

PR #156, opened 2025-01-17
Clearer byte range definition in `get_range` and `get_ranges`

kylebarron:
Closes https://github.com/developmentseed/obstore/issues/155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.4.0] -
+
+### Breaking changes :wrench:
+
+- `get_range`, `get_range_async`, `get_ranges`, and `get_ranges_async` now require named parameters for `start`, `end`, and `length` to make the semantics of the range request fully explicit. by @kylebarron in https://github.com/developmentseed/obstore/pull/156
+
 ## [0.3.0] - 2025-01-16
 
 ### New Features :magic_wand:
--- a/obstore/python/obstore/_get.pyi
+++ b/obstore/python/obstore/_get.pyi
@@ -261,7 +261,14 @@ async def get_async(
     Refer to the documentation for [get][obstore.get].
     """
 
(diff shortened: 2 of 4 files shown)


kylebarron:
cc @maxrjones A Rust match on two optional integers should be _extremely_ cheap, and much cheaper than matching Python objects. I guess it really depends on the Python overhead of translating keyword arguments into rust. In terms of the Rust overhead of the option match, I'd expect that should be on the order of nanoseconds. Maybe any additional Python overhead would add up to a microsecond extra?

@maxrjones Would you be able to rerun your zarr benchmark from this branch?

maxrjones:
> @maxrjones Would you be able to rerun your zarr benchmark from this branch?

Happy to. Can you provide guidance on how to install the crate + bindings from git? usually I'd use `uv add "obstore @ git+https://github.com/developmentseed/obstore@0ce55be897206d9867c6a1170be95c0c1ffc4d32"` but that raises `Package metadata name "test-env" does not match given name "obstore"`.

kylebarron:
I think you need `subdirectory=obstore` https://stackoverflow.com/a/19516714

kylebarron:
Just checked and 

```
pip install git+https://github.com/developmentseed/obstore@0ce55be897206d9867c6a1170be95c0c1ffc4d32#subdirectory=obstore
```

worked for me

maxrjones (approved):
Sorry for the delay, took me a second to be able to use compiled code on the VEDA Hub and I wanted to run the benchmarks in us-west-2. Anyways, the new rate for loading a ~3.7GB array with Zarr is ~0.36% slower which IMO is acceptable.

maxrjones:
I used a instance with half as many cores so it was ~2x that time for each. If you'd be concerned about that relative difference, if real, I could record the individual timings to determine if there's a statistically significant difference.

kylebarron:
Nah, that's good. Thanks for checking that benchmark!

PR #157, opened 2025-01-17
bytes pr draft

jessekrubin:
Currently copy-pasta-ed the `pyo3-bytes` impl and will PR into there once I get thoughts from @kylebarron.

Something is super screwy in the `decode` function 

Tests are in `ry/tests/test_bytes.py` and a a bit crude

Key Additions:

- slicing in `__getItem__`
- `hex` function
- attempt at `decode` 
- the `__repr__` string returns a string that should eval to an equal `Bytes` obj (i dont think it is technically identical to how python's builtin bytes do stringifying but it works (see tests))


kylebarron:
Thanks! I'm about to go on vacation for a few days, so I probably won't have time to review this until Wednesday

jessekrubin:
Happy vacation!

jessekrubin:
@kylebarron

added the tests also figured out the `decode` problemo. was ye-olde-cstr` needing that null char... 

I commented out the check for `0` length buf, so as to make it more python `bytes` like -- that way you can `obstore.Bytes()` and/or `obstore.Bytes(b'')`

should I add myself to the authors/contribs thing on the cargo.toml? (either way is fine for me)

pyo3-bytes/src/bytes.rs
```diff
@@ -396,3 +594,36 @@ impl<'py> FromPyObject<'py> for PyBytesWrapper {
         Ok(Self(Some(buffer)))
     }
 }
+
+/// Returns a string like `b"asdf\n\0"` for the given byte slice.
+fn python_bytes_repr(data: &[u8]) -> String {
```
kylebarron, line 599: 1. If we implement `std::fmt::Debug` for `PyBytes`, we don't have to first materialize the entire buffer to a string and then _again_ copy it into the 
    ```rs
    format!("Bytes({})", python_bytes_repr(self.0.as_ref()))
    ```
	line above. This can be really costly for large buffers (see https://github.com/georust/wkt/pull/124 for a related example of a big speedup by avoiding repeated string copies, if you're curious)

2. Is there no Rust API that handles this for us? It looks like the Debug for `Bytes` does this? https://github.com/tokio-rs/bytes/blob/71824b095c4150b3af0776ac158795c00ff9d53f/src/fmt/debug.rs#L6-L40

pyo3-bytes/src/bytes.rs
```diff
@@ -93,17 +95,24 @@ impl PyBytes {
     // By setting the argument to PyBytes, this means that any buffer-protocol object is supported
     // here, since it will use the FromPyObject impl.
     #[new]
+    #[pyo3(signature = (buf = PyBytes(Bytes::new())))]
```
kylebarron, line 98: This is the signature shown to Python, right? This isn't valid Python code.

kylebarron:
Can we split up these edits into smaller PRs, please? I think that will make it easier to review, because I think different sections of this PR are at different levels of stability. E.g. the `__repr__`, especially if we can use the `Bytes` debug impl, should be pretty stable, while the decoding at a glance looks less so (I'm hesitant to use any APIs from `pyo3::ffi` other than for the buffer protocol)

kylebarron:
> I commented out the check for `0` length buf, so as to make it more python `bytes` like -- that way you can `obstore.Bytes()` and/or `obstore.Bytes(b'')`

That sounds good!

> should I add myself to the authors/contribs thing on the cargo.toml? (either way is fine for me)

Sure, you can add yourself to the authors field of `pyo3-bytes/Cargo.toml`

pyo3-bytes/src/bytes.rs
```diff
@@ -396,3 +594,36 @@ impl<'py> FromPyObject<'py> for PyBytesWrapper {
         Ok(Self(Some(buffer)))
     }
 }
+
+/// Returns a string like `b"asdf\n\0"` for the given byte slice.
+fn python_bytes_repr(data: &[u8]) -> String {
+    let mut out = String::from("b\"");
+    for &byte in data {
+        match byte {
+            b'\\' => out.push_str("\\\\"),
```
kylebarron, line 603: Also, instead of adding these extra `\`, we could just put `br` in the repr. I.e. these are equal:
```py
br"\n" == b"\\n"
```
So the repr can be
```
Bytes(br"\n")
```

pyo3-bytes/src/bytes.rs
jessekrubin: > 1. If we implement `std::fmt::Debug` for `PyBytes`, we don't have to first materialize the entire buffer to a string and then _again_ copy it into the
>    ```rust
>    format!("Bytes({})", python_bytes_repr(self.0.as_ref()))
>    ```
>    
>    line above. This can be really costly for large buffers (see [Implement geo-traits for writing to WKT & perf improvement georust/wkt#124](https://github.com/georust/wkt/pull/124) for a related example of a big speedup by avoiding repeated string copies, if you're curious)

makes sense to me! I can change later today I think. 

Do you mean `Display`? or `Debug`? 

> 2. Is there no Rust API that handles this for us? It looks like the Debug for `Bytes` does this? https://github.com/tokio-rs/bytes/blob/71824b095c4150b3af0776ac158795c00ff9d53f/src/fmt/debug.rs#L6-L40

HAHAHAHA I did not see that...

pyo3-bytes/src/bytes.rs
jessekrubin: No I don't think it is. IIRC `__text_signature__` is?

pyo3-bytes/src/bytes.rs
jessekrubin: wont that not matter if we use the bytes debug impl?

jessekrubin:
> Can we split up these edits into smaller PRs, please? I think that will make it easier to review, because I think different sections of this PR are at different levels of stability. E.g. the `__repr__`, especially if we can use the `Bytes` debug impl, should be pretty stable, while the decoding at a glance looks less so (I'm hesitant to use any APIs from `pyo3::ffi` other than for the buffer protocol)
> 
> Let's start with just the `__repr__` add in one PR

I shall see what I can do after work! the decode does appear to work well now though, once I figured out the null-terminated c-str. 

Clarification. do you mean impl Debug? or Display?

jessekrubin:
@kylebarron see #173 for ya repr

pyo3-bytes/src/bytes.rs
kylebarron: We need the `r` raw string to ensure the bytes debug repr is correct

kylebarron:
I'm going to close this and we can iterate on future smaller PRs, like we did for the `__repr__`

jessekrubin:
@kylebarron bueno! I have been v busy at work, but can PR stuff in soon.

Issue #164, opened 2025-01-23
Support obstore as storage for df.to_parquet()


machichima:
Currently, using obstore's `AsyncFsspecStore` for `df.to_parquet() ` does not work (no error but also no files uploaded). It seems like we only need to add `write()` method in the `BufferedFileSimple` class and use the `obstore.put()` there.

https://github.com/developmentseed/obstore/blob/b40d59b4e060ba4fd3dc69468b3ba7da1149758e/obstore/python/obstore/fsspec.py#L177-L186

I would like ask if there is any existing progress on it? If not, I can help working on this.

kylebarron:
A PR for this would be great!

martindurant:
Does obstore support multi-part uploads, or would whole files need to be buffered in memory before a call to put()?

martindurant:
> You can pass a sync or async iterator to put

That's not much use from the file API, I'm afraid - you need to be able to write, send some data and return; so the iterator would be the wrong way around.

machichima:
Hi @kylebarron ,

I tried it out using the iterator. However, just like what @martindurant metioned, the whole file will be bufferd in the memory, which is not ideal. Probably the Rust part needs to be updated to support this.

martindurant:
At the moment, obstore doesn't implement a file API at all, I guess this is in the name. As well as buffering in memory, you could also use fsspec's `simplecache` to buffer to local file and then upload that - also not entirely a perfect solution.

kylebarron:
> At the moment, obstore doesn't implement a file API at all

obstore implements a file API for reading but it's not yet implemented for writing. See https://developmentseed.org/obstore/v0.3.0/api/file/

I'm open to having a similar API for writing that wraps https://docs.rs/object_store/latest/object_store/buffered/struct.BufWriter.html

kylebarron:
> so the iterator would be the wrong way around.

I'm not sure I fully understand this use case, but if the iterator is the wrong way around, then maybe instead of a "push-based" iterator, you could pass in a "pull-based" file-like object? That's already supported.

machichima:
> I'm not sure I fully understand this use case, but if the iterator is the wrong way around, then maybe instead of a "push-based" iterator, you could pass in a "pull-based" file-like object? That's already supported.

Hi,
The problem here is that `to_parquet()` will call `write()` multiple times, each time it will provide a chunk of data (bytes). So the thing we can pass to `obstore.put()` is either bytes or iterator, and not file-like object.
[...]

PR #165, opened 2025-01-27
[FEAT] support df.to_parquet and df.read_parquet()

machichima:
Add write() function for `BufferedFileSimple` used whan calling fsspec.open().

https://github.com/developmentseed/obstore/blob/b40d59b4e060ba4fd3dc69468b3ba7da1149758e/obstore/python/obstore/fsspec.py#L177-L186

Related to issue #164


obstore/python/obstore/fsspec.py
```diff
@@ -176,13 +177,49 @@ async def _ls(self, path, detail=True, **kwargs):
...
+        if self._writer.closed():
+            return
+        self._upload_chunk(final=True)
+        self._writer.close()
+
+
+class BufferedFileRead(fsspec.spec.AbstractBufferedFile):
     def __init__(self, fs, path, mode="rb", **kwargs):
-        if mode != "rb":
-            raise ValueError("Only 'rb' mode is currently supported")
```
kylebarron, line 185: Why remove this?

obstore/python/obstore/fsspec.py
```diff
@@ -176,13 +177,49 @@ async def _ls(self, path, detail=True, **kwargs):
 
     def _open(self, path, mode="rb", **kwargs):
         """Return raw bytes-mode file-like from the file-system"""
-        return BufferedFileSimple(self, path, mode, **kwargs)
```
kylebarron, line 179: can we assert that the mode is either rb or wb?

obstore/python/obstore/fsspec.py
```diff
@@ -28,6 +28,7 @@
 import fsspec.spec
 
 import obstore as obs
+from obstore import open_reader, open_writer
```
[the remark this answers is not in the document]
machichima: Ah yes, I used this as original BufferedFileSimple inherit from `AbstractBufferedFile`, which is not async. I will change it to `AbstractAsyncStreamedFile` and use async method

obstore/python/obstore/fsspec.py
machichima: I am thinking of adding the assert in `_open()`, so I think we don't need to assert it again in the `BufferedFileRead` and `BufferedFileWrite`?

https://github.com/developmentseed/obstore/blob/75d373433d214c1db049e6ad5d0ed93aada6fd8b/obstore/python/obstore/fsspec.py#L178-L183

obstore/python/obstore/fsspec.py
machichima: Sure! I'll add it

machichima:
I found that there's also some bug in checking if parquet file exists in info() so I rename the title

obstore/python/obstore/fsspec.py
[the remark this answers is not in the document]
machichima: It seems like `df.to_parquet()` and `df.read_parquet()` does not support async, as I got error like `RuntimeWarning: coroutine 'AbstractAsyncStreamedFile.write' was never awaited` if trying to change to async

machichima:
Hi @kylebarron ,

I am wondering about the test here. Originally, `fs.info("dir")` for directory will raise file not found error, which cause error in using `df.to_parquet()`. After fixing it, the line `fs.cat("dir", recursive=True)`  will raise `FileNotFoundError` for "dir" as `fs.info("dir")` has no error so "dir" will be processed. 

https://github.com/developmentseed/obstore/blob/428a66dffa60abae32592a110789b18205b5bfcd/tests/test_fsspec.py#L47-L59

Should I try to make its output as `{"dir/test1": b"test data1", "dir/test2": b"test data2"}` here? Which requires to override `_cat()` in fsspec as follow

```python
    async def _cat(
        self, path, recursive=False, on_error="raise", batch_size=None, **kwargs
    ):
        paths = await self._expand_path(path, recursive=recursive)
        coros = [self._cat_file(path, **kwargs) for path in paths if not self._isdir(path)]   # ignore dir for cat_file
        batch_size = batch_size or self.batch_size
```

Refer to fsspec, it simply gives `FileNotFoundError` when doing so. Maybe we can just remove this line or make it assert if FileNotFound raise?

kylebarron:
@martindurant wrote that test and is obviously more familiar with fsspec than I am... @martindurant do you have any suggestions here?

martindurant:
Keep pinging me until I have a chance to look at this :)

obstore/python/obstore/fsspec.py
```diff
@@ -197,12 +224,15 @@ async def _ls(self, path, detail=True, **kwargs):
...
 
-    def _open(self, path, mode="rb", **kwargs):
+    def _open(self, path: str, mode="rb", **kwargs):
         """Return raw bytes-mode file-like from the file-system"""
         assert mode in (
             "rb",
             "wb",
         ), f"Only 'rb' and 'wb' mode is currently supported, got: {mode}"
+
+        _, path = self._split_path(path)
```
kylebarron, line 234: We should assert that the bucket of the path matches the bucket of the store.

machichima: Hi @kylebarron ,
I run `print(dir(store))` and it outputs:

```sh
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'from_env', 'from_session', 'from_url']
```
It seems like obstore does not provide any attribute to get the bucket name from the store instance. Maybe we should add something like this in Rust?

```rust
#[getter]
fn bucket(&self) -> Option<String> {
    self.config.bucket.clone()
}

```

kylebarron: For the present release I think we could add a private method `_bucket`, however bucket won't always be defined, such as when the store is created by url.

obstore/python/obstore/fsspec.py
```diff
@@ -148,17 +176,36 @@ async def _get_file(self, rpath, lpath, **kwargs):
...
+                "name": head["path"],
+                "size": head["size"],
+                "type": "directory" if head["path"].endswith("/") else "file",
+                # Implementation-specific keys
+                "e_tag": head["e_tag"],
+                "last_modified": head["last_modified"],
+                "version": head["version"],
+            }
+        except FileNotFoundError:
+            # try ls, refer to the info implementation in fsspec
```
martindurant, line 192: Why does this PR need the extra code? Are you trying open() with globs? I don't know the details on head_async, whether it might already achieve this.

obstore/python/obstore/fsspec.py
```diff
@@ -177,15 +224,58 @@ async def _ls(self, path, detail=True, **kwargs):
...
             return sorted([object["path"] for object in objects] + prefs)
 
-    def _open(self, path, mode="rb", **kwargs):
+    def _open(self, path: str, mode="rb", **kwargs):
         """Return raw bytes-mode file-like from the file-system"""
-        return BufferedFileSimple(self, path, mode, **kwargs)
+        assert mode in (
+            "rb",
+            "wb",
+        ), f"Only 'rb' and 'wb' mode is currently supported, got: {mode}"
```
martindurant, line 232:
```suggestion
        if mode not in {"rb", "wb"}:
            raise ValueError(f"Only 'rb' and 'wb' mode is currently supported, got: {mode}")
```
I know @kylebarron said "assert", but I think this is clearer. It is unusual to see `assert` outside of tests.

obstore/python/obstore/fsspec.py
```diff
@@ -177,15 +224,58 @@ async def _ls(self, path, detail=True, **kwargs):
...
+            self.buffer.seek(0)
+            self._writer.write(self.buffer.read())
+            # flush all the data in buffer when closing
+            if final:
+                self._writer.flush()
+            return True
+        else:
+            return False
+
+    def close(self):
```
martindurant, line 267: Should also set `self.closed = True`

obstore/python/obstore/fsspec.py
```diff
@@ -177,15 +224,58 @@ async def _ls(self, path, detail=True, **kwargs):
...
+        if mode == "wb":
+            return BufferedFileWrite(self, path, mode, **kwargs)
+        if mode == "rb":
+            return BufferedFileRead(self, path, mode, **kwargs)
+
+
+class BufferedFileWrite(fsspec.spec.AbstractBufferedFile):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._writer = open_writer(self.fs.store, self.path)
```
martindurant, line 245: This shouldn't be here, since you also have it in _initiate_upload
```suggestion
```

obstore/python/obstore/fsspec.py
```diff
@@ -177,15 +224,58 @@ async def _ls(self, path, detail=True, **kwargs):
...
+        """
+        Called by AbstractBufferedFile flusH() on the first flush
+        """
+        self._writer = open_writer(self.fs.store, self.path)
+
+    def _upload_chunk(self, final=False):
+        """
+        Called every time fsspec flushes the write buffer
+        """
+        if self.buffer and len(self.buffer.getbuffer()) > 0:
```
martindurant, line 257: It shouldn't be possible to get here without this condition being True

martindurant (commented):
I made some comments on the code as it stands. 

However, the outstanding issues is: how to construct these instances via fsspec.open(). It would mean
- registering each of the expected protocols (s3, gs, ab) to override the fsspec default ones. Perhaps a top-level function in obstore would do this explicitly (I wouldn't do it implicitly on import).
- writing a _get_kwargs_from_urls to create the right obstore instance for the given path(s), including the bucket. This would also be a way to stash the value of the bucket, for later asserting the paths are right.

The alternative way, annoying for the user, would be to explicitly pass a premade instance with filesystem= (sometimes fs=) to the given loading function.

obstore/python/obstore/fsspec.py
machichima: For the code that stores the parquet as: `file.csv/00000`, `file.csv/00001`, ...etc, when reading the `file.csv/` from s3, the `info()` will give `FileNotFoundError`. As I known, s3's folder is not an object but a prefix, which cause this error from happending. So I add the code if getting `FileNotFoundError` in `head_async` to solve it

machichima:
> I made some comments on the code as it stands.
> 
> However, the outstanding issues is: how to construct these instances via fsspec.open(). It would mean
> 
> * registering each of the expected protocols (s3, gs, ab) to override the fsspec default ones. Perhaps a top-level function in obstore would do this explicitly (I wouldn't do it implicitly on import).
> * writing a _get_kwargs_from_urls to create the right obstore instance for the given path(s), including the bucket. This would also be a way to stash the value of the bucket, for later asserting the paths are right.
> 
> The alternative way, annoying for the user, would be to explicitly pass a premade instance with filesystem= (sometimes fs=) to the given loading function.

Hi @martindurant ,

I've opened a new [draft PR](https://github.com/developmentseed/obstore/pull/198) to ensure consistency in how instances are constructed across methods. My goal is to align the usage with fsspec.

With this PR, obstore can be registered as an fsspec storage backend using:

```python
fsspec.register_implementation("s3", S3FsspecStore)
```

The bucket is extracted from the file path and used as a cache key when creating obstore objects. Here's an example usage that I would like to achieve:

```python
fsspec.register_implementation("s3", S3FsspecStore)
fs: AsyncFsspecStore = fsspec.filesystem(
    "s3",
    config={
        "endpoint": "http://localhost:30002",
        "access_key_id": "minio",
        "secret_access_key": "miniostorage",
        "virtual_hosted_style_request": True,  # path contain bucket name
    },
    client_options={"timeout": "99999s", "allow_http": "true"},
    retry_config={
        "max_retries": 2,
        "backoff": {
            "base": 2,
            "init_backoff": timedelta(seconds=2),
        },
    },
)

fs.cat_file("my-s3-bucket/test.txt")

```

Does this align with your expectations? Please let me know if you have any suggestions!
Thanks!

obstore/python/obstore/fsspec.py
```diff
@@ -81,6 +83,32 @@ def __init__(
             *args, asynchronous=asynchronous, loop=loop, batch_size=batch_size
         )
 
+    def _split_path(self, path: str) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
machichima: I named it as `_split_path` to align with the naming in s3fs [here](https://github.com/fsspec/s3fs/blob/023aecf00b5c6243ff5f8a016dac8b6af3913c6b/s3fs/core.py#L443), as this function is doing the same thing as s3fs's `split_path`

obstore/python/obstore/fsspec.py
```diff
@@ -81,6 +83,32 @@ def __init__(
...
+            # no bucket name in path
+            return "", path
+
+        if "/" not in path:
+            return path, ""
+        else:
+            path_li = path.split("/")
+            bucket = path_li[0]
+            file_path = "/".join(path_li[1:])
+            return (bucket, file_path)
```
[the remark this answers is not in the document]
machichima: I added following code in https://github.com/developmentseed/obstore/pull/198 but haven't sync here yet. This can solve the :// issue

```python
if path.startswith(self.protocol + "://"):
    path = path[len(self.protocol) + 3 :]
elif path.startswith(self.protocol + "::"):
    path = path[len(self.protocol) + 2 :]
path = path.rstrip("/")
```

obstore/python/obstore/fsspec.py
```diff
@@ -177,15 +224,58 @@ async def _ls(self, path, detail=True, **kwargs):
...
+            "wb",
+        ), f"Only 'rb' and 'wb' mode is currently supported, got: {mode}"
+
+        _, path = self._split_path(path)
+
+        if mode == "wb":
+            return BufferedFileWrite(self, path, mode, **kwargs)
+        if mode == "rb":
+            return BufferedFileRead(self, path, mode, **kwargs)
+
```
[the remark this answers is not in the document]
machichima: Just added the exception in else here

obstore/python/obstore/fsspec.py
machichima: I think we can remove `self.buffer` here, but for `len(self.buffer.getbuffer()) > 0`, I think we might need this when `flush(force=True)` when closing? Which does not ensure buffer contains data. See: https://github.com/fsspec/filesystem_spec/blob/f30bc759f30327dfb499f37e967648f175750fac/fsspec/spec.py#L2041

obstore/python/obstore/fsspec.py
martindurant: OK, the same old "is it a folder" problem - I am well familiar with this.

obstore/python/obstore/fsspec.py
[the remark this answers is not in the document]
martindurant: fsspec also allows "s3a" for the same thing - I don't know if you want to allow that.

Also, I don't know how "::" can appear here - it exists to join path elements, not protocol to path.

obstore/python/obstore/fsspec.py
kylebarron, line 192: Do we need to duplicate this from upstream? Can we just call `self._ls`?

obstore/python/obstore/fsspec.py
kylebarron: This is fixed in https://github.com/developmentseed/obstore/pull/210, so you can access bucket information out.

martindurant:
> I think having obstore.fsspec.register, even if that function is a one-liner that wraps fsspec.register

Exactly what I was thinking - the user can call register themselves as int he example above, but it would be useful to provide a utility function that knows what to register, so the user only needs to call one thing once.

obstore/python/obstore/fsspec.py
machichima: I'll try this out

obstore/python/obstore/fsspec.py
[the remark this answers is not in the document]
machichima: Oh I actually copied this from fsspec and thought obstore may do the similar stuff. It looks like obstore does not need this, I'll take them out then.

obstore/python/obstore/fsspec.py
machichima: It works, I updated here:

https://github.com/developmentseed/obstore/blob/ff5d6bd09cc37cd786e88c28bf095d673149dee6/obstore/python/obstore/fsspec.py#L191-L194

machichima:
I will continue on this once [this PR](https://github.com/developmentseed/obstore/pull/198) is merge, so that we can use the new way to construct the obstore insntance in `open()` too

obstore/python/obstore/fsspec.py
```diff
@@ -220,6 +220,14 @@ def _construct_store(self, bucket: str) -> ObjectStore:
...
         )
 
+    def split_path(self, path: str) -> tuple[str, str]:
+        """Public method to split a path into bucket and path components."""
+        return self._split_path(path)
+
+    def construct_store(self, bucket: str) -> ObjectStore:
+        """Public method to construct a store for the given bucket."""
+        return self._construct_store(bucket)
+
```
kylebarron, line 230: What's the rationale for making these public?

obstore/python/obstore/fsspec.py
```diff
@@ -542,5 +737,6 @@ def _register(protocol: str, *, asynchronous: bool) -> None:
                 "asynchronous": asynchronous,
             },  # Assign protocol dynamically
         ),
-        clobber=False,
+        # Override any existing implementations of the same protocol
+        clobber=True,
```
kylebarron, line 748: Note, I changed this, because I think we want to override any of the default fsspec protocols

kylebarron:
I pushed a couple updates here:

- _Always_ use the underlying Rust `BufferedFile` and `BufferedWriter` where possible, and use the provided fsspec methods as little as possible. 
    - Override the provided `read`, `tell`, `seek`, `readline`, `readlines`, `flush`, `write` methods
    - Use the underlying rust reader for `readline` and `readlines`
- Added typing overloads for the `BufferedFile` `__init__` for either readers/writers.
- Minimized what we set on `self`. We don't need to set `fs`, `store`, or `path` on `self`.

kylebarron:
Thanks for starting this!

obstore/python/obstore/fsspec.py
```diff
@@ -478,25 +490,71 @@ def __init__(
...
         if length < 0:
-            data = self.fs.cat_file(self.path, self.loc, self.size)
-            self.loc = self.size
-        else:
-            data = self.fs.cat_file(self.path, self.loc, self.loc + length)
-            self.loc += length
-        return data
+            length = self.size - self.loc
+
+        self._reader.seek(self.loc)
```
kylebarron: We shouldn't seek on every underlying read call if we don't need to. Seeking discards the internal buffer kept by rust

(Forgot to post this comment earlier)

machichima:
Thank you so much!

PR #173, opened 2025-01-28
Add `__repr__` to `Bytes` class

jessekrubin:
stripped down version of #157 

- JUST bytes `__repr__()`


pyo3-bytes/src/bytes.rs
```diff
@@ -93,6 +93,7 @@ impl PyBytes {
     // By setting the argument to PyBytes, this means that any buffer-protocol object is supported
     // here, since it will use the FromPyObject impl.
     #[new]
+    #[pyo3(signature = (buf = PyBytes(Bytes::new())))]
```
kylebarron, line 96: Continuing from https://github.com/developmentseed/obstore/pull/157#discussion_r1932733891

You're right it doesn't go through to the python text signature, but that means it's currently empty.

image

We should probably include `text_signature` as well, with `buf` set to `b""`

pyo3-bytes/src/bytes.rs
```diff
@@ -396,3 +401,39 @@ impl<'py> FromPyObject<'py> for PyBytesWrapper {
         Ok(Self(Some(buffer)))
     }
 }
+
+/// Implement Debug for PyBytes
+///
+/// Don't use the `bytes::Bytes` Debug impl, b/c it doesn't _look_
```
kylebarron, line 407: I don't think it has to look exactly the same, as long as it's accurate. I'd rather reuse the upstream implementation. And if we use `br""` then the upstream debug string should fully round trip.

tests/test_bytes.py
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+
+from obstore import Bytes
+
+
+@pytest.mark.skip(reason="Not implemented *yet*")
+def test_decode() -> None:
+    py_bytes = b"asdf"
+    ry_bytes = Bytes(py_bytes)
+    assert ry_bytes.decode() == "asdf"
```
kylebarron, line 12: Can we remove this from this PR, because it doesn't implement decoding?

tests/test_bytes.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# TODO: figure out if accepts empty bytes (i jessekrubin think it should if it is trying to be `bytes`)
```
kylebarron, line 15: Remove todo?

pyo3-bytes/src/bytes.rs
```diff
@@ -327,9 +330,13 @@ impl PyBytes {
 
 impl<'py> FromPyObject<'py> for PyBytes {
     fn extract_bound(ob: &Bound<'py, PyAny>) -> PyResult<Self> {
-        let buffer = ob.extract::<PyBytesWrapper>()?;
-        let bytes = Bytes::from_owner(buffer);
-        Ok(Self(bytes))
+        if ob.is_none() {
```
kylebarron, line 333: We shouldn't automatically accept `None` for PyBytes. Downstream users can explicitly opt-in to accepting `None` by using `Option<PyBytes>` in their type

pyo3-bytes/src/bytes.rs
jessekrubin: I don't follow.... the signature does go through as `(buf)` -- you are saying you want to explicitly set it to `(buf = b"")`?


```
In [12]: inspect.signature(ry.Bytes)
Out[12]: <Signature (buf)>
```

pyo3-bytes/src/bytes.rs
jessekrubin: boom done

tests/test_bytes.py
jessekrubin: boom

tests/test_bytes.py
jessekrubin: boom

pyo3-bytes/src/bytes.rs
jessekrubin: I dont think it is possible to do the repr with `br""` feel free to try. 

- the current `Bytes` Debug impl outputs `b"..."` which precludes using `br`
- again, I dont think it is possible as it would lead to invalid evals
- the current impl DOES round trip and the tests handle all possible `u8` (AFAIK) 
- the python style with ascii chars appears to be a shorter string

jessekrubin: The problem is with '\' -- can't br with it as far as I know.

pyo3-bytes/src/bytes.rs
kylebarron: Yes, setting the text_signature to `b""` makes it clear that the default value of `buf` is an empty byte string.

kylebarron: image

So the `text_signature` is valid input

kylebarron:
Awesome, thanks!

jessekrubin:
Anytime! I'll pr in the slicing soon.

Lmk if/when you publish.

Issue #90, opened 2024-11-17
Single store API for all backends, where the url chooses which backend


kylebarron:
E.g. `new_store(url)` which could load from env, or `new_store(url, kwargs)`

daviewales:
It would be amazing to be able to do something like:

``` python
import obstore as obs
from obstore.store import PrefixStore

store = PrefixStore.from_url('abfs://container@account.dfs.core.windows.net/path')
blobs = obs.list(store)
print(blobs)
```

i.e. Initialise a new PrefixStore from a URL in one line.
I've been hunting everywhere for the simplest method to list the blobs at an arbitrary location, but there's nothing as simple as this.

Even more amazing would be to simplify it at follows:

``` python
import obstore as obs
blobs = obs.list('abfs://container@account.dfs.core.windows.net/path')
print(blobs)
```

`fsspec` request something like the following:

``` python
from urllib.parse import urlparse
import fsspec

source_url = 'abfs://container@account.dfs.core.windows.net/path'
source = urlparse(source_url)
source_account = source.hostname.split(".", 1)[0] if source.hostname else None

fs = fsspec.filesystem(
    source.scheme,
    account_name=source_account,
    anon=False,
)
blobs = fs.ls(source_url)
print(blobs)
```

This is not terrible, but it requires me to manually extract all the components from the URL.

kylebarron:
I spent a lot of time today on the API design here, and I have https://github.com/developmentseed/obstore/pull/185 and https://github.com/developmentseed/obstore/pull/179, which together should make it really easy to create a store by URL.

Here's an example with a test bucket of mine
<img src="https://github.com/user-attachments/assets/7d18a05b-1826-42c6-99aa-256ef6e98526" height="400"/>

`new_store` checks the scheme of the URL and passes it to the relevant class' `from_url` classmethod. And with https://github.com/developmentseed/obstore/pull/185 you won't need to use a `PrefixStore` anymore; the stores will automatically handle a prefix. And `from_url` automatically checks for any environment variables, so you can automatically load data even from a private bucket.

(feedback welcomed on the name of `new_store`)

kylebarron:
> Even more amazing would be to simplify it at follows:
> ```
> import obstore as obs
> blobs = obs.list('abfs://container@account.dfs.core.windows.net/path')
> print(blobs)
> ```

[...]

PR #179, opened 2025-01-30
Add top-level `from_url` function

kylebarron:
Closes https://github.com/developmentseed/obstore/issues/90

Todo:

- [ ] ~~Note: will not read from the environment.~~ Now that we're passing to e.g. `S3Store::from_url` it _should_ actually read from the environment! (thanks to https://github.com/developmentseed/obstore/pull/189) (note: we need to test this!!)
- [ ] Add test for reading from the env.
- [ ] ~~Note: Cannot be used for `sign`~~
- [x] Add to docs

~~We may want to vendor `parse_url_opts` ([which is pretty small](https://github.com/apache/arrow-rs/blob/3bf29a2c7474e59722d885cd11fafd0dca13a28e/object_store/src/parse.rs#L142-L214)), so that we can return the Python classes directly, rather than returning a `PrefixStore`. At least if we switch to storing a prefix store. #184~~


pyo3-object_store/src/url.rs
```diff
@@ -0,0 +1,28 @@
+use pyo3::exceptions::PyValueError;
+use pyo3::pybacked::PyBackedStr;
+use pyo3::types::PyAnyMethods;
+use pyo3::FromPyObject;
+use url::Url;
+
+/// A wrapper around [`url::Url`] that implements [`FromPyObject`].
+pub struct PyUrl(Url);
```
kylebarron, line 8: Note: we already have this in `signer.rs`

pyo3-object_store/src/simple.rs
```diff
@@ -0,0 +1,60 @@
...
+    let url = url.into_inner();
+    let (scheme, _) = ObjectStoreScheme::parse(&url).map_err(object_store::Error::from)?;
+    let (store, path) = match scheme {
+        ObjectStoreScheme::AmazonS3 => {
+            let combined = crate::aws::combine_config_kwargs(
+                config.map(|x| x.extract()).transpose()?,
+                kwargs.map(|x| x.extract()).transpose()?,
+            )?
+            .unwrap_or_default();
+            parse_url_opts(&url, combined.into_inner())?
```
kylebarron, line 29: Once #185 is merged, we can switch this to just 
```
PyS3Store::from_url(url, config.map(|x| x.extract()).transpose()?, None, None, kwargs.map(|x| x.extract()).transpose()?)
```
Then we can make `combind_config_kwargs` mod private again.

And we don't even have to vendor `parse_url_opts` I think!

kylebarron:
This PR is waiting on #185, which will enable the prefix handling to work out of the box

pyo3-object_store/src/simple.rs
```diff
@@ -0,0 +1,99 @@
...
+) -> PyObjectStoreResult<PyObject> {
+    let url = url.into_inner();
+    let (scheme, _) = ObjectStoreScheme::parse(&url).map_err(object_store::Error::from)?;
+    match scheme {
+        ObjectStoreScheme::AmazonS3 => {
+            let store = PyS3Store::from_url(
+                &PyType::new::<PyS3Store>(py),
+                url.as_str(),
+                config.map(|x| x.extract()).transpose()?,
+                None,
```
kylebarron, line 36: todo: pass retry config and http client config

kylebarron:
@maxrjones what do you think about naming this
```
obstore.store.from_url
```
?

Then it's also less surprising that it delegates to `from_url` of each individual store.

maxrjones:
> @maxrjones what do you think about naming this
> 
> ```
> obstore.store.from_url
> ```
> 
> ?
> 
> Then it's also less surprising that it delegates to `from_url` of each individual store.

I like it!

Issue #125, opened 2025-01-08
Support for pickling


norlandrhagen:
Hi there,

@jbusecke and I were playing around with using obstore to read netcdf files into Xarray as part of apache-beam data ingestion pipeline. This requires the serialization of objects that are passed between stages. We ran into an error when beam was trying to pickle a dataset. We reproduced the behavior **without** beam.

Anyhow, wondering if it's possible to pickle obstore-backed Xarray datasets or if this is a limitation that we don't understand.

**Create the example dataset**
```python

import xarray as xr 
ds = xr.tutorial.open_dataset('air_temperature')
ds.to_netcdf('air.nc')

```
**This works with fsspec** 

```python
import fsspec 
import xarray as xr 
import cloudpickle 

fs = fsspec.filesystem('local')
ds_fsspec = xr.open_dataset(fs.open('air.nc'), engine='h5netcdf', chunks={})

with open('fsspec.pkl', 'wb') as f:
    cloudpickle.dump(ds_fsspec, f)

```
**This fails with obstore's fsspec wrapper**

```python


from obstore.fsspec import AsyncFsspecStore
from obstore.store import LocalStore 
from pathlib import Path
import xarray as xr 

store = LocalStore(prefix=Path("."))
fss = AsyncFsspecStore(store)
ds = xr.open_dataset(fss.open('air.nc'), engine='h5netcdf', chunks={})


import cloudpickle 
with open('obstore.pkl', 'wb') as f:
    cloudpickle.dump(ds, f)

```

`TypeError: cannot pickle 'builtins.LocalStore' object`

traceback:

<details>

```python
{
	"name": "TypeError",
	"message": "cannot pickle 'builtins.LocalStore' object",
	"stack": "---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 14
     12 import cloudpickle 
     13 with open('obstore.pkl', 'wb') as f:
---> 14     cloudpickle.dump(ds, f)

File ~/miniforge3/envs/obstore/lib/python3.12/site-packages/cloudpickle/cloudpickle.py:1511, in dump(obj, file, protocol, buffer_callback)
   1498 def dump(obj, file, protocol=None, buffer_callback=None):
   1499     \"\"\"Serialize obj as bytes streamed into file
   1500 
   1501     protocol defaults to cloudpickle.DEFAULT_PROTOCOL which is an alias to
   (...)
   1509     next).
   1510     \"\"\"
[...]

PR #184, opened 2025-01-30
proof of concept: Pickle support 

kylebarron:
[Pickle support](https://github.com/developmentseed/obstore/issues/125) is important but hard to implement. 

## Background

There are two ways to implement pickle support. 

1. Implementing [`__getstate__`](https://docs.python.org/3/library/pickle.html#object.__getstate__) and [`__setstate__`](https://docs.python.org/3/library/pickle.html#object.__setstate__). The return value of `__getstate__` can be pretty much anything I think, and that gets passed back into `__setstate__` to 
2. Implementing a constructor (tagged in Rust with `#[new]`) and [`__getnewargs_ex__`](https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__). `__getnewargs_ex__` must return a tuple of `(args: tuple, kwargs: dict)`, which can be passed to the `#[new]` constructor. This can be simpler when you already have a `#[new]` implemented and when the parameters into that `#[new]` function are easily serializable. However it might require  (I wonder if there's a way to recursively pickle things?)

We can't extract the configuration out from a "finished" `object_store` instance like an [`AmazonS3`](https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3.html). We could extract some configuration out of a builder instance like [`AmazonS3Builder`](https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html), but then _every time we use that class_ we'd have to call `build()` on the Rust side, which would probably significantly hurt performance. Therefore, the only(?) possible way to implement pickle support inside obstore is to persist the store configuration separately inside the `#[pyclass]`.

## Implementation

- Store configuration information _a second time_ in the `#[pyclass]` of each store. (It's already implicitly stored by the underlying rust `ObjectStore` instance.
- Implement `__getnewargs_ex__`, returning the parameters passed to `#[new]`
- Implement `IntoPyObject` for each configuration object: the store-specific config, client options, and retry options

## Benefits

- It should be relatively straightforward to implement for any of the raw stores: `S3Store`, `GCSStore`, `AzureStore`, `LocalStore`, and `MemoryStore`.
- We're already validating the `PyAmazonS3Config`, `PyClientOptions`, and `PyRetryConfig`, so it isn't that much extra work just to store those on the Rust class. So we can serialize them to Python objects in `__getnewargs_ex__` and then have Python automatically pass them to `#[new]`.
- Using `__getnewargs_ex__` means we don't need to add serde support; we can use `IntoPyObject` and `FromPyObject` for all (de)serialization and only have a single code path.
- We can persist all of the store-specific config, the client options, and the retry config, so the pickled instances should be _exactly_ the same as the original instances.
- Supports any of the builder classmethods, e.g. `from_env`, `from_url`, etc,
- Most of the time, storing configuration information should be just a few strings. So ideally it'll increase memory usage only slightly, and won't affect runtime performance otherwise (assuming you reuse a store instead of creating a new one each time).
- Since we don't allow the store classes to be mutated after creation from Python, there's no risk of the two copies of the configuration getting out of sync.

## Drawbacks

- Unclear how to support middleware, including `PrefixStore`, because those have to support an _arbitrary_ wrapped object. Is there a way to recursively pickle and unpickle the thing it's wrapping?
	- If we can't find a way to support pickling of arbitrary middleware, we could alternatively use a `PrefixStore` internally and automatically inside an `S3Store`, `GCSStore`, `AzureStore` (NOTE: we should maybe benchmark the overhead a `PrefixStore` causes, in case it's something we don't want to force on everyone? Well, if the `S3Store` stored an arbitrary `Arc<dyn ObjectStore>` then we could prefix when asked for and not prefix when not asked for, but maybe that would conflict with signing, if that requires a raw `object_store::AmazonS3` instance?). Alternatively, we could create the `PrefixStore` on demand, since it [looks virtually free](https://github.com/apache/arrow-rs/blob/3bf29a2c7474e59722d885cd11fafd0dca13a28e/object_store/src/prefix.rs#L44-L49).
- `MemoryStore` won't copy its state, but that's unavoidable I think. (Should be documented)

## Todo

- [ ] Implement `IntoPyObject` for `ClientOptions` and `RetryConfig`.
- [ ] Docs page on pickling and any gotchas (e.g. MemoryStore and PrefixStore limitations).

(Note, we should probably add this issue content to a dev doc)

Closes https://github.com/developmentseed/obstore/issues/125

### DIFF
--- a/pyo3-object_store/src/aws.rs
+++ b/pyo3-object_store/src/aws.rs
@@ -3,10 +3,10 @@ use std::str::FromStr;
 use std::sync::Arc;
 
 use object_store::aws::{AmazonS3, AmazonS3Builder, AmazonS3ConfigKey};
-use pyo3::intern;
 use pyo3::prelude::*;
 use pyo3::pybacked::PyBackedStr;
-use pyo3::types::PyType;
+use pyo3::types::{PyDict, PyString, PyTuple, PyType};
+use pyo3::{intern, IntoPyObjectExt};
 
 use crate::client::PyClientOptions;
 use crate::config::PyConfigValue;
@@ -14,19 +14,25 @@ use crate::error::{ObstoreError, PyObjectStoreError, PyObjectStoreResult};
 use crate::retry::PyRetryConfig;
 
 /// A Python-facing wrapper around an [`AmazonS3`].
-#[pyclass(name = "S3Store", frozen)]
-pub struct PyS3Store(Arc<AmazonS3>);
+#[pyclass(name = "S3Store", module = "obstore.store", frozen)]
+pub struct PyS3Store {
+    store: Arc<AmazonS3>,
+    bucket: String,
+    config: Option<PyAmazonS3Config>,
+    client_options: Option<PyClientOptions>,
+    retry_config: Option<PyRetryConfig>,
+}
 
 impl AsRef<Arc<AmazonS3>> for PyS3Store {
     fn as_ref(&self) -> &Arc<AmazonS3> {
-        &self.0
+        &self.store
     }
 }
 
 impl PyS3Store {
     /// Consume self and return the underlying [`AmazonS3`].
     pub fn into_inner(self) -> Arc<AmazonS3> {
-        self.0
+        self.store
     }
 }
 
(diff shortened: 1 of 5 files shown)


kylebarron:
Trying out one way forward in https://github.com/developmentseed/obstore/pull/185

kylebarron:
Superseded by #185

PR #185, opened 2025-01-31
Support pickling & always manage store prefix

kylebarron:
See #184 for background

### Changes

- Remove separate PrefixStore. Always manage prefix for AWS, GCP, Azure.

TODO:

- [x] Document pickle support. Any middlewares in the future won't be able to be pickled.
- [x] Add dev doc to repo

Closes https://github.com/developmentseed/obstore/issues/125


maxrjones:
Thanks for your work on this @kylebarron!

How do I get the prefix to be recognized when using `S3Store.from_url`:

e.g.,
```Python
import obstore as obs

bucket = "nasa-veda-scratch"
path = "zarr-obstore-test/max/zarr-v3/test-era5-v3-919"

# Open store with obstore
s3store = S3Store.from_url(f"s3://{bucket}/{path}")
stream = obs.list(s3store, chunk_size=10)
for list_result in stream:
    print(list_result[0])
    break
```

Currently returns:
```
{'path': 'icechunk-test/max/met/manifests/4RCPNN1YNY6REGPDAQBG', 'last_modified': datetime.datetime(2025, 1, 27, 21, 8, 48, tzinfo=datetime.timezone.utc), 'size': 214, 'e_tag': '"5622f6e96f657cfc1089e5c5b53f923c"', 'version': None}
```

kylebarron:
Hmm this might be a bug in my prefix handling

obstore/python/obstore/store/_aws.pyi
```diff
@@ -521,6 +521,7 @@ class S3Store:
         self,
         bucket: str | None = None,
         *,
+        prefix: str | None = None,
```
maxrjones, line 524: does this signature change also need to happen in `_azure.pyi` and `_gcs.pyi`?

obstore/python/obstore/store/_aws.pyi
```diff
@@ -544,8 +546,9 @@ class S3Store:
     def from_session(
         cls,
         session: boto3.session.Session | botocore.session.Session,
-        bucket: str,
+        bucket: str | None = None,
         *,
+        prefix: str | None = None,
```
maxrjones, line 551: same questions as above about whether this change needs to be copied to the other store classes?

maxrjones (commented):
Is it possible to implement `__getnewargs_ex__` for LocalStore?

kylebarron:
@geospatial-jeff benchmarked this (from the 0.4.0-beta.1 tag) as _maybe_ 3% slower than the previous 0.3.0, but that could be partially variation between tests, and would be worth it anyways for these changes.

PR #198, opened 2025-02-03
[FEAT] Create obstore store in fsspec on demand

machichima:
Construct the obstore store instance on demand in fsspec when calling methods. This allows automatic store creation for reads/writes across different buckets, aligning usage with fsspec conventions


kylebarron:
Would there be one fsspec instance per cloud provider? So if you wanted to use s3 and gcs you'd make two separate instances?

obstore/python/obstore/fsspec.py
```diff
@@ -45,6 +47,9 @@ def __init__(
         self,
         store: obs.store.ObjectStore,
         *args,
+        config: dict[str, Any] = {},
```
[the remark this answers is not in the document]
machichima: I use store here for deciding the store Interface (whether it is S3Store, GCSStore, ...), so that in AsyncFsspecStore we don't need to decide the interface based on the protocol. 

Maybe there's a better way of deciding the store interface?

```python
obstore_fs: AsyncFsspecStore = fsspec.filesystem(
    "s3",
    store=S3Store,
    config={
        "endpoint": "http://localhost:30002",
        "access_key_id": "minio",
        "secret_access_key": "miniostorage",
        "virtual_hosted_style_request": True,  # path contain bucket name
    },
    client_options={"timeout": "99999s", "allow_http": "true"},
    retry_config={
        "max_retries": 2,
        "backoff": {
            "base": 2,
            "init_backoff": timedelta(seconds=2),
        },
    },
)

```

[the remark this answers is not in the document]
machichima: I'll have a look at the typing later on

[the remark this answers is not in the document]
kylebarron: Oh that's confusing because store is the type of the class and not an instance. 

We should be able to use the from_url top level function directly here?

kylebarron:
It would be nice to take out the `store` arg and use `from_url` directly. `from_url` will automatically construct the correct store based on the url protocol

kylebarron:
The latest PRs allow you to access the `config` back out of a store, which may be useful to you? You can validate that you already have a store in your cache for a specific bucket

machichima:
> Is it true that a single fsspec class can't be associated with more than one protocol? E.g. Azure has three different protocols `abfs`, `adlfs` and `az`, but it [looks like adlfs exports three separate classes](https://github.com/fsspec/adlfs/blob/adb9c53b74a0d420625b86dd00fbe615b43201d2/adlfs/__init__.py#L4).

I think we can if those protocol refer to the same object instancel. s3fs do have 2 protocols, see: https://github.com/fsspec/s3fs/blob/023aecf00b5c6243ff5f8a016dac8b6af3913c6b/s3fs/core.py#L277

I think abfs, adlfs, and az have different implementation so that they exports classes

obstore/python/obstore/fsspec.py
```diff
@@ -104,6 +104,12 @@ def _split_path(self, path: str) -> Tuple[str, str]:
             # no bucket name in path
             return "", path
 
+        if path.startswith(self.protocol + "://"):
+            path = path[len(self.protocol) + 3 :]
+        elif path.startswith(self.protocol + "::"):
```
kylebarron, line 109: Can you give an example of where `::` would show up? Or maybe even add it to a test?

kylebarron:
> I think we can if those protocols refer to the same object instance. s3fs do have 2 protocols ("s3", "s3a"), see: [fsspec/s3fs@`023aecf`/s3fs/core.py#L277](https://github.com/fsspec/s3fs/blob/023aecf00b5c6243ff5f8a016dac8b6af3913c6b/s3fs/core.py#L277)

Oh cool! That seems to indicate that we could have a single class that defines supported protocols as:
```py
    protocol = ("s3", "s3a", "gs", "az", "abfs", "adlfs")
```

Because the fsspec class used for each is the same? It's just custom kwargs that would need to be passed down for each?

machichima:
> Oh cool! That seems to indicate that we could have a single class that defines supported protocols as:
> 
> ```python
>     protocol = ("s3", "s3a", "gs", "az", "abfs", "adlfs")
> ```
> 
> Because the fsspec class used for each is the same? It's just custom kwargs that would need to be passed down for each?

I don't think we can put all the protocols together into a class, as when using `fsspec.register_implementation("s3", AsyncFsspecStore)`, fsspec wouldn't tell `AsyncFsspecStore` what the protocol is, so that when constructing store instance, we cannot get the protocol

https://github.com/developmentseed/obstore/blob/6614906c689c2fedd51482feba7f1dc01ad3c10d/obstore/python/obstore/fsspec.py#L122-L124


I think the better way is to create `obstore.fsspec.register("protocol")`, that wraps around the `fsspec.register` and directly set the protocol for `AsyncFsspecStore` (like what mentioned in [this comment](https://github.com/developmentseed/obstore/pull/165#issuecomment-2640220852)), then we do not need more classes. Let me have a try.

obstore/python/obstore/fsspec.py
```diff
@@ -104,6 +104,12 @@ def _split_path(self, path: str) -> Tuple[str, str]:
             # no bucket name in path
             return "", path
 
+        if path.startswith(self.protocol + "://"):
```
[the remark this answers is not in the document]
machichima: It will not always be `s3://mybucket/path/to/file`, but may be without protocol like `mybucket/path/to/file`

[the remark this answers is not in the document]
machichima: If we do not need to support "::" format, then we can use this like below for separating protocol and path, also ensure the protocol is same as expected

```python
        res = urlparse(path)
        if res.scheme:
            if res.scheme != self.protocol:
                raise ValueError(f"Expect protocol to be {self.protocol}. Got {res.schema}")
            path = res.path

```

obstore/python/obstore/fsspec.py
```diff
@@ -104,11 +105,11 @@ def _split_path(self, path: str) -> Tuple[str, str]:
...
 
-        if path.startswith(self.protocol + "://"):
-            path = path[len(self.protocol) + 3 :]
-        elif path.startswith(self.protocol + "::"):
-            path = path[len(self.protocol) + 2 :]
-        path = path.rstrip("/")
+        res = urlparse(path)
+        if res.scheme:
+            if res.scheme != self.protocol:
+                raise ValueError(f"Expect protocol to be {self.protocol}. Got {res.schema}")
```
kylebarron, line 111: The linter should catch this too. You have `schema`

kylebarron:
I did a quick look through your PR; it's really good progress but a few thoughts:

- There are a bunch of cases where `bucket, path = self._split_path(path)` doesn't work because `path` is not in scope. E.g. in `_cp_file` where `path1` and `path2` are in scope
- in `_cp_file` we need to validate that the bucket of the source and destination paths are the same
- We need some tests for edits that happen in this PR
- It's not clear how `BufferedFileSimple` works, because that subclasses from an upstream `fsspec.spec.AbstractBufferedFile` but doesn't touch `obstore` apis at all
- If you don't already, I'd highly suggest using a linter like https://docs.astral.sh/ruff/ in your editor, so that you can catch some of these issue before hitting CI

machichima:
> * If you don't already, I'd highly suggest using a linter like https://docs.astral.sh/ruff/ in your editor, so that you can catch some of these issue before hitting CI

Thanks for the suggestion! I just added ruff linter and remove error for path. I also add the check for validating the two bucket name from two path are the same.

> * It's not clear how `BufferedFileSimple` works, because that subclasses from an upstream `fsspec.spec.AbstractBufferedFile` but doesn't touch `obstore` apis at all

For `BufferedFileSimple`, when `self.fs.cat_file()` is called, it will direct to the `_cat_file()` function in `AsyncFsspecStore`.

> * We need some tests for edits that happen in this PR

Yes! I will update the test in the next few days

machichima:
Hi @kylebarron , 

I would like to confirm if changing the output for `_ls` to the path with bucket name is fine here? 

The problem I faced is that when I call `find()`, it will get into `_find()` in fsspec, which will then call `_walk()` (see [here](https://github.com/fsspec/filesystem_spec/blob/f30bc759f30327dfb499f37e967648f175750fac/fsspec/asyn.py#L852)). Then in `_walk()`, `_ls` is called (see [here](https://github.com/fsspec/filesystem_spec/blob/f30bc759f30327dfb499f37e967648f175750fac/fsspec/asyn.py#L718)). But as `_walk()` will run recurrsively if the path is directory (see [here](https://github.com/fsspec/filesystem_spec/blob/f30bc759f30327dfb499f37e967648f175750fac/fsspec/asyn.py#L755-L759)), if `_ls` give the path without bucket name, then when walk is called recurrsively by the previous result, `_ls` will get the path without bucket name, which cause error.

For example, call `_walk("bucket/")` first, and the `_ls` in `_walk` give "dir/", then walk will be called again with `_walk("dir/")`, which gives error

kylebarron:
> I would like to confirm if changing the output for `_ls` to the path with bucket name is fine here?

I think that's a question for @martindurant. I don't know what's standard for fsspec

kylebarron:
@machichima I merged a PR that updates our use of [ruff](https://docs.astral.sh/ruff), a fast Python linter. As part of this, there were some minor updates to the existing fsspec code. Would you be able to merge in main?

I also wanted to have ruff merged so that we could run the lints on this PR's changes.

kylebarron:
> > I would like to confirm if changing the output for `_ls` to the path with bucket name is fine here?
> 
> I think that's a question for @martindurant. I don't know what's standard for fsspec
> 
> It appears that based on `s3fs` behavior, the bucket name is always returned in list results:
> 
> ```python
> import s3fs
> 
> fs = s3fs.S3FileSystem(anon=True)
> fs.ls("sentinel-cogs")
> # ['sentinel-cogs/sentinel-s2-l2a-cogs']
> ```

It would be great if we could have integration tests with s3fs. We could reuse the existing test setup we have in `test_fsspec.py`, but then validate that we get the same results for all methods across both the obstore-backed store and the s3fs implementation.

martindurant:
Yes, fsspec expects that the paths returned by ls() are the same form as you can then use in further open/get/put etc operations. Since the input paths contain the bucket (and optional protocol), the output of ls() should too.

The alternative would be, to have the filesystem instance *not* expect the bucket in the input. You could still get this to work by extracting the bucket once (as passed in fsspec.open/url_to_fs). This is essentially what the "prefix" filesystem does:
```
In [10]: fs, path = fsspec.url_to_fs("dir::file:///Users")

In [11]: path
Out[11]: ''

In [12]: fs.ls("", detail=False) # contents of /Users
Out[12]: ['moxie', '.localized', 'Shared', 'mdurant']
```

machichima:
> @machichima I merged a PR that updates our use of [ruff](https://docs.astral.sh/ruff), a fast Python linter. As part of this, there were some minor updates to the existing fsspec code. Would you be able to merge in main?
> 
> I also wanted to have ruff merged so that we could run the lints on this PR's changes.

Sure, I'll merge and solve the conflicts tomorrow

kylebarron:
> Sure, I'll merge and solve the conflicts tomorrow

Let me know if you need any help; I added quite a few lints

machichima:
Hi @kylebarron ,
I just fixed the conflicts and update the code based on linter. Note that I add some more ruff ignore into `pyproject.toml`, please have a look to see if those are alright.

Thanks!

pyproject.toml
```diff
@@ -48,12 +50,18 @@ ignore = [
     "ANN204", # Missing return type annotation for special method
     "E501",   # Line too long
 ]
+"*.py" = [
+    "E501",   # Line too long
+    "FBT001", # Boolean positional argument in function definition (should be keyword-only)
+    "FBT002", # Boolean default positional argument in function definition (should be keyword-only)
+]
```
kylebarron, line 57: These should be scoped only to the `fsspec.py` file

pyproject.toml
```diff
@@ -4,7 +4,9 @@ version = "0.1.0"
 description = "Add your description here"
 readme = "README.md"
 requires-python = ">=3.11"
-dependencies = []
+dependencies = [
+    "cachetools>=5.5.1",
```
kylebarron, line 8: I don't want to add a dependency on any external project. If we _really had to_, we could use it as an optional dependency only within the fsspec integration, but I'd prefer to avoid any external dependency at all.

obstore/python/obstore/fsspec.py
```diff
@@ -152,79 +153,97 @@ def __init__(
...
+
+    def _construct_store(self, bucket: str) -> ObjectStore:
+        if bucket not in self._store_cache:
+            self._store_cache[bucket] = from_url(
+                url=f"{self.protocol}://{bucket}",
+                config=self.config,
+                client_options=self.client_options,
+                retry_config=self.retry_config or None,
+            )
+        return self._store_cache[bucket]
```
kylebarron, line 199: Isn't this doing exactly the same as the previous function iteration? This should be exactly the same as 
```py
@lru_cache(maxsize=10)
def _construct_store(self, bucket: str):
```
Either way, it's a cache using `bucket` as a cache key.

pyproject.toml
kylebarron: I pushed a quick fix to this: [`897beb0` (#198)](https://github.com/developmentseed/obstore/pull/198/commits/897beb07b84d5fda85776551015ae4ff09408a37)

obstore/python/obstore/fsspec.py
machichima: This triggered the B019 warning: https://docs.astral.sh/ruff/rules/cached-instance-method/, which will "retain a reference to the instance, preventing it from being garbage collected"

In the newest commit ([5b87c46](https://github.com/developmentseed/obstore/pull/198/commits/5b87c468a43799ddba9b200f685b326591e72329)) I used the method mentioned here: https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-instance-methods-release-object for using the weak reference. Which do not need external package.

kylebarron: Ah, well that's a good linter then, because I never would've caught that myself 🙂

kylebarron: I'm curious about this solution: https://stackoverflow.com/a/68550238

Apart from possibly being simpler, it would also let the user pass in the number of store instances to cache in the `__init__`. (ref above https://github.com/developmentseed/obstore/pull/198#discussion_r1939567445) Whereas with the current decorator approach I don't know if it's possible for a user to specify how big this cache should be?

kylebarron:
> Do I also need to add integration test in this PR?

I think this is an unequivocal improvement as-is. We can leave integration tests out of this PR and then come back to them in the next PR.

obstore/python/obstore/fsspec.py
machichima: In the current approach, we can set the max size for cache by adding the class attribute. However, I do think the solution you refer to is better. I will try this out.

Besides, I am thinking of if user called `fsspec.filesystem` more than once, which will create multiple filesystem instances. But the approaches we mentioned create cache for each instance, not class. There might be a situation that they create two filesystems with same config like so:

```python
fs_1: AsyncFsspecStore = fsspec.filesystem(
    "s3",
    config={...},
)
fs_2: AsyncFsspecStore = fsspec.filesystem(
    "s3",
    config={...},
)
```

And both run:

```python
fs_1.get("bucket/path")
fs_2.get("bucket/path")
```

Then `fs_2` will still instantiate the obstore instance instead of reading from cache. Refer to: https://rednafi.com/python/lru_cache_on_methods/#the-self-dilemma

Maybe we should turn the `_construct_store` into class method and also make config, client_config, retry_config as key?

machichima:
> Do you have some examples of how this new API should be used

Yes, I am currently writing script to test those functions locally. I can share the script here once I finished :smile:

machichima:
The test error in the latest commit is caused by using cat() for directory, which also happens in s3fs

obstore/python/obstore/fsspec.py
kylebarron: > But the approaches we mentioned create cache for each instance, not class. There might be a situation that they create two filesystems with same config like so:
> 
> ```python
> fs_1: AsyncFsspecStore = fsspec.filesystem(
>     "s3",
>     config={...},
> )
> fs_2: AsyncFsspecStore = fsspec.filesystem(
>     "s3",
>     config={...},
> )
> ```

Given that we're only caching instances by bucket name, I think it's beneficial for now to manage the cache as local to the instance. 

Besides, is there any reason for a user to create `fs_1` and `fs_2` from the same config? We can just document that they should reuse an existing `fsspec` instance.

kylebarron:
> The test error in the latest commit is caused by using cat() for directory, which also happens in s3fs

Are you saying that there's a bug here that doesn't exist in s3fs? Or that we have the same behavior as s3fs??

obstore/python/obstore/fsspec.py
machichima: > Besides, is there any reason for a user to create fs_1 and fs_2 from the same config? We can just document that they should reuse an existing fsspec instance.

It is useful to use fsspec.filesystem() as "get or create filesystem" so that we do not need to check if the filesystem is already instantiate or not (NOTE: to do this, we should set `cachable = True`, which I did it in the newer commit)

```python
def func1(path):
       file_system = self.get_filesystem_for_path(path)
       file_system.get(...)

def func2(path):
       file_system = self.get_filesystem_for_path(path)
       file_system.put(...)

def get_filesystem_for_path(path):
        protocol = get_protocol(path)

        if protocol == "s3":
            s3kwargs = s3_setup_args(_data_config.s3, anonymous=anonymous)
            s3kwargs.update(kwargs)
            return fsspec.filesystem(protocol, **s3kwargs) 
       ...


```

machichima:
> Are you saying that there's a bug here that doesn't exist in s3fs? Or that we have the same behavior as s3fs?

I think it's a bug in s3fs and we do behave the same. This is caused by the way of dealing with folder in s3 (see this comment: https://github.com/developmentseed/obstore/pull/165#discussion_r1944707081). 

The fastest way of dealing this is to override the `_cat()` (see this comment: https://github.com/developmentseed/obstore/pull/165#issuecomment-2624561876), but this might not be the best way of dealing this.

Future-Outlier (approved):
lgtm

kylebarron:
Tests are failing; would you be able to fix that?

machichima:
The failing tests are XFail ones, I don't know why github now catching those. Should I take them out?

machichima:
> Tests are failing; would you be able to fix that?

Hi @kylebarron 
I just add the code to clean up after each test. The OOM error is due to using register("s3") multiple times that create multiple classes, those classes then stuck in the memory. I set the `clobber=False` in register to prevent this.

obstore/python/obstore/fsspec.py
```diff
@@ -296,3 +468,66 @@ def read(self, length: int = -1) -> Any:
             data = self.fs.cat_file(self.path, self.loc, self.loc + length)
             self.loc += length
         return data
+
+
+def register(protocol: str | list[str], asynchronous: bool = False) -> None:
```
kylebarron, line 473:
```suggestion
def register(protocol: str | list[str], *, asynchronous: bool = False) -> None:
```
I'm a big proponent of having a small number of positional parameters. This also allows us to make `asynchronous` a positional argument in the future without it being breaking

obstore/python/obstore/fsspec.py
```diff
@@ -296,3 +468,66 @@ def read(self, length: int = -1) -> Any:
...
+    if isinstance(protocol, list):
+        # Ensure all elements are strings
+        if not all(isinstance(p, str) for p in protocol):
+            raise TypeError("All protocols in the list must be strings.")
+        # Ensure no empty strings in the list
+        if not all(p for p in protocol):
+            raise ValueError("Protocol names in the list must be non-empty strings.")
+
+        for p in protocol:
+            register(p)
```
kylebarron, line 519: We need to pass down the value of `asynchronous`

obstore/python/obstore/fsspec.py
```diff
@@ -296,3 +468,66 @@ def read(self, length: int = -1) -> Any:
...
+        >>> register(["gcs", "abfs"])  # Registers both "gcs" and "abfs"
+
+    Notes:
+        - Each protocol gets a dynamically generated subclass named
+          `AsyncFsspecStore_<protocol>`.
+        - This avoids modifying the original AsyncFsspecStore class.
+
+    """
+    # Ensure protocol is of type str or list
+    if not isinstance(protocol, str | list):
```
kylebarron, line 498: Does this syntax work in 3.9?

Regardless, we should allow any iterable input, not just a `List`

obstore/python/obstore/fsspec.py
```diff
@@ -296,3 +467,50 @@ def read(self, length: int = -1) -> Any:
             data = self.fs.cat_file(self.path, self.loc, self.loc + length)
             self.loc += length
         return data
+
+
+def register(protocol: str | Iterable[str], *, asynchronous: bool = False) -> None:
```
kylebarron, line 472: This can be a follow-up PR, but we should support calling `register` with no arguments, which registers for all supported backends.

kylebarron:
I have to run, but I can do a final review later today or tomorrow

kylebarron:
I tried a basic list operation and couldn't get it to work with this PR.

A successful list with s3fs:
```py
import s3fs
fs = s3fs.S3FileSystem(anon=True)
path = "s3://sentinel-cogs/sentinel-s2-l2a-cogs/12/S/UF/2022/6/S2B_12SUF_20220609_0_L2A"
fs.ls(path)
```
prints
```py
['sentinel-cogs/sentinel-s2-l2a-cogs/12/S/UF/2022/6/S2B_12SUF_20220609_0_L2A/AOT.tif',
 'sentinel-cogs/sentinel-s2-l2a-cogs/12/S/UF/2022/6/S2B_12SUF_20220609_0_L2A/B01.tif',
...
```

Trying to do this with this PR:

```py
# Neither of these two work
store = AsyncFsspecStore(config={"skip_signature": True})
store = AsyncFsspecStore("s3", config={"skip_signature": True})
path = "s3://sentinel-cogs/sentinel-s2-l2a-cogs/12/S/UF/2022/6/S2B_12SUF_20220609_0_L2A"
store.ls(path)
```

raises with
```
GenericError: Generic URL error: Unable to recognise URL "abstract://"
```

You're checking `self.protocol` in multiple places but it's never getting set.

tests/conftest.py
```diff
@@ -49,3 +49,13 @@ def s3_store(s3: str):
...
     )
+
+
+@pytest.fixture
+def s3_store_config(s3: str):
+    return {
+        "AWS_ENDPOINT_URL": s3,
+        "AWS_REGION": "us-east-1",
+        "AWS_SKIP_SIGNATURE": "True",
+        "AWS_ALLOW_HTTP": "true",
```
kylebarron, line 60:
```suggestion
        "AWS_SKIP_SIGNATURE": True,
        "AWS_ALLOW_HTTP": True,
```

tests/conftest.py
```diff
@@ -49,3 +49,13 @@ def s3_store(s3: str):
         aws_skip_signature=True,
         client_options={"allow_http": True},
     )
+
+
+@pytest.fixture
+def s3_store_config(s3: str):
```
kylebarron, line 55:
```suggestion
def s3_store_config(s3: str) -> S3ConfigInput:
```

obstore/python/obstore/fsspec.py
```diff
@@ -59,27 +72,48 @@ class AsyncFsspecStore(fsspec.asyn.AsyncFileSystem):
...
         Keyword Args:
             asynchronous: Set to `True` if this instance is meant to be be called using
                 the fsspec async API. This should only be set to true when running
                 within a coroutine.
+            max_cache_size (int, optional): The maximum number of stores the cache
+                should keep. A cached store is kept internally for each bucket name.
+                Defaults to 10.
             loop: since both fsspec/python and tokio/rust may be using loops, this
                 should be kept `None` for now, and will not be used.
             batch_size: some operations on many files will batch their requests; if you
```
kylebarron, line 119: This example below is no longer valid. Can you update it?

obstore/python/obstore/fsspec.py
```diff
@@ -59,27 +72,48 @@ class AsyncFsspecStore(fsspec.asyn.AsyncFileSystem):
     fsspec-style object.
     """
 
-    cachable = False
+    # https://github.com/fsspec/filesystem_spec/blob/56054c0a30ceedab4c0e6a0f7e429666773baf6d/docs/source/features.rst#instance-caching
+    cachable = True
 
-    def __init__(
+    def __init__(  # noqa: PLR0913
```
kylebarron, line 78: It seems you want a `protocol` parameter here, which sets the value of `protocol` onto `self`?

obstore/python/obstore/fsspec.py
```diff
@@ -98,19 +132,83 @@ def __init__(
...
+            (bucket name, file path inside the bucket)
+
+        Examples:
+            >>> split_path("s3://mybucket/path/to/file")
+            ['mybucket', 'path/to/file']
+
+        """
+        protocol_with_bucket = ["s3", "s3a", "gcs", "gs", "abfs", "https", "http"]
+
+        if self.protocol not in protocol_with_bucket:
```
kylebarron, line 165: What are examples of protocols that we support that are not any of the above?

obstore/python/obstore/fsspec.py
```diff
@@ -98,19 +132,83 @@ def __init__(
...
+        if res.scheme:
+            # path is in url format
+            if res.scheme != self.protocol:
+                err_msg = f"Expect protocol to be {self.protocol}. Got {res.scheme}"
+                raise ValueError(err_msg)
+            return (res.netloc, res.path)
+
+        # path not in url format
+        if "/" not in path:
+            return path, ""
```
kylebarron, line 179: Aren't you returning a tuple of `(bucket, file_path)`? Then returning `path, ""` doesn't make sense.

obstore/python/obstore/fsspec.py
```diff
@@ -98,19 +132,83 @@ def __init__(
...
+        path_li = path.split("/")
+        bucket = path_li[0]
+        file_path = "/".join(path_li[1:])
```
kylebarron, line 182: Much simpler than this is to split on only the first `/` character:
```
path = "bucket/path/to/file.txt"
path.split("/", 1)
# ["bucket", "path/to/file.txt"]
```
Then you don't need to split and rejoin the path

obstore/python/obstore/fsspec.py
kylebarron: It seems we'd want to error in this case? There's no way for us to infer the bucket to use.

As it stands you're searching through the string twice, once here and again below in `path.split`.

Instead, you can call `path.split("/", 1)` once. If the result is a list of length 1, then you know a `"/"` wasn't in the path, and then you can error.

obstore/python/obstore/fsspec.py
```diff
@@ -98,19 +132,83 @@ def __init__(
...
+        self._construct_store = lru_cache(maxsize=max_cache_size)(self._construct_store)
+
         super().__init__(
             *args,
             asynchronous=asynchronous,
             loop=loop,
             batch_size=batch_size,
         )
 
+    def _split_path(self, path: str) -> tuple[str, str]:
```
kylebarron, line 149: The only thing this function uses from `self` is `self.protocol`. Let's move `_split_path` into global scope, and then we can test `_split_path` specifically from the test file.

We should validate that we can split the path both for URLs with a protocol and for "paths" without the protocol.

obstore/python/obstore/fsspec.py
```diff
@@ -192,31 +324,62 @@ async def _put_file(
...
+                "type": "directory" if head["path"].endswith("/") else "file",
+                # Implementation-specific keys
+                "e_tag": head["e_tag"],
+                "last_modified": head["last_modified"],
+                "version": head["version"],
+            }
+        except FileNotFoundError:
+            # use info in fsspec.AbstractFileSystem
+            loop = asyncio.get_running_loop()
+            return await loop.run_in_executor(None, super().info, path, **_kwargs)
```
kylebarron, line 378: Why are you calling `super().info`? You can call `super()._info`, which is async, and not need to touch the running event loop at all.

obstore/python/obstore/fsspec.py
```diff
@@ -192,31 +324,62 @@ async def _put_file(
...
+
+    @staticmethod
+    def _fill_bucket_name(path: str, bucket: str) -> str:
```
kylebarron, line 381: This is only used in two places. Can we just copy the f-string above and delete this helper function?

machichima:
> You're checking `self.protocol` in multiple places but it's never getting set.

This is because the protocol is set in class variable in register. I did this because `fsspec.filesystem(protocol, config)` will not pass the protocal to the `AsyncFsspecStore` class (see [here](https://github.com/fsspec/filesystem_spec/blob/56054c0a30ceedab4c0e6a0f7e429666773baf6d/fsspec/registry.py#L296-L310)). I took a look at how s3fs did, they simply set the protocol in class variable (see [here](https://github.com/fsspec/s3fs/blob/b4d3fa89f1accbfc2186f1d42d07be15bc9df101/s3fs/core.py#L277))

https://github.com/developmentseed/obstore/blob/5ea2ba81e5df25977f9de3ed0b295fd67ed2aefc/obstore/python/obstore/fsspec.py#L521-L532

To make your example work, you can do:

```python
AsyncFsspecStore.protocol = "s3"
store = AsyncFsspecStore(config={"skip_signature": True, "region": "us-west-2"})
path = "s3://sentinel-cogs/sentinel-s2-l2a-cogs/12/S/UF/2022/6/S2B_12SUF_20220609_0_L2A"
print(store.ls(path))
```

I think we we should add protocol as the __init__ parameter so that user can also set protocol when directly usinng `AsyncFsspecStore`

obstore/python/obstore/fsspec.py
machichima: I referred here: https://developmentseed.org/obstore/latest/api/store/#obstore.store.from_url, where options "memory" and "file" will not have bucket name

obstore/python/obstore/fsspec.py
machichima: I did this for case like: `fs.ls("bucket_name")`, in this case, only bucket name are provided. I'll change using `path.split("/", 1)`

kylebarron:
Thank you! Ping me again when it's ready for a review!

obstore/python/obstore/fsspec.py
machichima: We can still test the _split_path without moving it to global scope. I think it is better to keep this in class so that we do not need to pass `self.protocol` into it from all methods

https://github.com/developmentseed/obstore/blob/a212996420d7ff7f1aca92bb1286aa0af72dd259/tests/test_fsspec.py#L200-L223

obstore/python/obstore/fsspec.py
machichima: This is because `_info` in `AsyncFileSystem` is `NotImplemented`, see https://github.com/fsspec/filesystem_spec/blob/56054c0a30ceedab4c0e6a0f7e429666773baf6d/fsspec/asyn.py#L701-L702

machichima:
Hi @kylebarron 

I fixed all the reviews above. Please have a look and let me know if any more fix needed. Thanks

tests/conftest.py
kylebarron: Correct, `allow_http` is a client option because it gets passed to the HTTP client, and is shared among all store implementations.

obstore/python/obstore/fsspec.py
kylebarron: If it's just those two, it's simpler to check if the protocol _is_ in that, rather than not in the converse

PR #199, opened 2025-02-03
Standardize on obstore's NotFoundError

kylebarron:
## Available PR templates

### DIFF
--- a/pyo3-object_store/src/error.rs
+++ b/pyo3-object_store/src/error.rs
@@ -1,7 +1,7 @@
 //! Contains the [`PyObjectStoreError`], the error enum returned by all fallible functions in this
 //! crate.
 
-use pyo3::exceptions::{PyFileNotFoundError, PyIOError, PyNotImplementedError, PyValueError};
+use pyo3::exceptions::{PyIOError, PyNotImplementedError, PyValueError};
 use pyo3::prelude::*;
 use pyo3::{create_exception, DowncastError};
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


ion-elgreco:
Also fine with keeping the Py FileNotFoundError, I think in delta-rs we also raise that one instead of a custom exception

kylebarron:
> Also fine with keeping the Py FileNotFoundError, I think in delta-rs we also raise that one instead of a custom exception

I'm debating this too... I'm leaning towards using the `obstore.exceptions.NotFoundError` so that it can subclass from a common base exception from `obstore.exceptions`?

ion-elgreco:
> > Also fine with keeping the Py FileNotFoundError, I think in delta-rs we also raise that one instead of a custom exception
> 
> I'm debating this too... I'm leaning towards using the `obstore.exceptions.NotFoundError` so that it can subclass from a common base exception from `obstore.exceptions`?

That might be better if you want to handle exceptions across obstore with the base exception.

kylebarron:
@martindurant How can we tell fsspec a list of allowed exceptions? Switching from `FileNotFoundError` to `obstore.exceptions.NotFoundError` is breaking the fsspec integration

martindurant:
Simplest would be to subclass

kylebarron:
I think one possible solution here is to make `NotFoundError` subclass from both `BaseError` and `FileNotFoundError`. However, it's unclear if multiple inheritance is possible in pyo3.

Issue #220, opened 2025-02-05
Add dagster-obstore to docs under integrations


kylebarron:
https://github.com/dagster-io/community-integrations/pull/107

ion-elgreco:
🚀

Issue #152, opened 2025-01-16
Development docs/guide


jessekrubin:
I was able to setup the repo for development and general editable stuff, but it wasn't obvious to me how to do so at first. 

I think a bare minimum development guide set of docs would be great and would be a boon for potential contributors. 

Pain points I experienced:
- what is the layout and where should I focus within the repo to get setup
- maturin/pyo3 WITH uv was a bit confusing at first
- where should I write my tests and dev scripts? 

Goal(s) (imo):
- provide minimal readme/docs/words describing how the repo is organized
- provide minimal readme for dev setup

I think these small things would be a nice improvement and I would love to help out w it but i don't (necessarily) think I am the person to do it(?)

kylebarron:
Yes, I agree, all projects should have a `DEVELOP.md` and it's an oversight that this one doesn't yet.

As a tip you can check the tests for the `uv` commands to run.

It's basically just
```
uv run maturin develop -m obstore/Cargo.toml
```
to build `obstore` and put it in your top-level venv.

And then
```
uv run mkdocs serve
```
to build and host the docs locally.

You can see how the develop.md looks for other similar  projects of mine:

https://github.com/kylebarron/arro3/blob/main/DEVELOP.md
https://github.com/geoarrow/geoarrow-rs/blob/a2cbc7d501ea4fcd4d5e8927d09bde08ccc34124/python/DEVELOP.md#L4
https://github.com/developmentseed/lonboard/blob/main/DEVELOP.md

PR #227, opened 2025-02-05
Add contributor, performance, integrations docs

kylebarron:
Closes https://github.com/developmentseed/obstore/issues/152, closes https://github.com/developmentseed/obstore/issues/220


kylebarron:
> My key pain point is that uv wants to build a release version all the time.

That happens when your python env is also the root of the python directory. See https://github.com/kylebarron/geo-index/blob/9ca258bf397250c46b52e0abaef47e29a5c71913/python/DEVELOP.md

PR #234, opened 2025-02-06
User-supplied credential callback

### LINKED ISSUE #269: Custom google auth via `google-auth`
Jotting down some notes from my exploration. This would become a callback for #234 

```py
# Sync
import google.auth
from google.auth.transport.requests import Request

credentials, project = google.auth.default()
request = Request()
credentials.refresh(request)
credentials.expiry
credentials.token_state


# Async

from google.auth._default_async import default_async
from google.auth.transport._aiohttp_requests import Request as AsyncRequest

async_request = AsyncRequest()
async_creds, project = default_async()
await async_creds.refresh(async_request)
async_creds.expiry
async_creds.token_state
async_creds.token
```

- `google.auth` user guide: https://google-auth.readthedocs.io/en/latest/user-guide.html
- [Polars code](https://github.com/pola-rs/polars/blob/1e16f141780579d97ae03f5ab39c36f9f652e675/py-polars/polars/io/cloud/credential_provider/_providers.py#L407-L423)
- `google.auth.transport`: https://google-auth.readthedocs.io/en/master/reference/google.auth.transport.html
- [`google.auth.transport` Async docs](https://googleapis.dev/python/google-auth/latest/reference/google.auth.transport._aiohttp_requests.html)

kylebarron:
There's a myriad number of ways to handle credentials for each of these stores, and I don't want to be implementing every last one of these. Luckily, `object_store` allows for [external credential providers](https://docs.rs/object_store/latest/object_store/trait.CredentialProvider.html), and we can allow users to implement their own totally custom authentication in Python!

This is a proof of concept that is tested as working with both a synchronous or asynchronous credential provider! Here are a couple examples:
```py
def credential_provider() -> S3Credential:
    session = boto3.Session(aws_access_key_id=aws_access_key_id, aws_secret_access_key=aws_secret_access_key, aws_session_token=aws_session_token)
    credentials = session.get_credentials().get_frozen_credentials()
    return {
        "access_key_id": credentials.access_key,
        "secret_access_key": credentials.secret_key,
        "token": credentials.token,
        "timeout": datetime.now(UTC) + timedelta(minutes=1000)
    }

store = S3Store("ds-wheels", credential_provider=credential_provider)
test = obs.list(store).collect()
```

```py
async def credential_provider() -> S3Credential:
    session = boto3.Session(aws_access_key_id=aws_access_key_id, aws_secret_access_key=aws_secret_access_key, aws_session_token=aws_session_token)
    credentials = session.get_credentials().get_frozen_credentials()
    return {
        "access_key_id": credentials.access_key,
        "secret_access_key": credentials.secret_key,
        "token": credentials.token,
        "timeout": datetime.now(UTC) + timedelta(minutes=1000)
    }

store = S3Store("ds-wheels", credential_provider=credential_provider)
await obs.list(store).collect_async()
```

### Notes:

- [x] Only try to refresh credentials (i.e. call the Python callback) when necessary. This includes handling the timeout datetime correctly. 
- [x] Add `credential_provider` to `obstore.store.from_url`
- [x] require tz set on expiry time
- [ ] Passing an async callback will hang if you use sync obstore APIs. You need to use async obstore APIs if you pass in an async callback. Is there a way for us to validate and error if the user passes in an async callback to a sync function? So we don't hang?
- [x] Example with at least S3, maybe [aws sts](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts/client/get_session_token.html), and maybe azure (planetary computer?) 
- [x] Use something like upstream's [`TokenCache`](https://github.com/apache/arrow-rs/blob/92cfd99e9ab4a6c54500ec65252027b9edf1ee55/object_store/src/client/token.rs#L35)
- [x] Update pickle "advanced" doc for whether this will work in pickle. It's ok for this not to work in pickle for now.
- ~~Expose from `obstore.store.google.auth` instead of `obstore.google.auth`?~~
- [x] Move all auth implementations to `obstore.auth`, which is a module we can keep expanding. So `obstore.auth.earthdata`, `obstore.auth.google`, `obstore.auth.boto3`,
- [x] ~~Create abstract base classes for auth callback?~~
    - Thinking of not having base classes to reinforce that protocols, not inheritance, is the intended usage. And type checkers enforce correctness of input. 
    - ~~These base classes should have some way to "declare" which store they're compatible with, as a simple validation when a class is passed into the provider callback~~
- [x] Update docs for "simple" and "complex" auth providers. Simple providers are just function callbacks; complex providers are class-based.
- [x] Naming: `AuthProvider` vs `CredentialProvider`? ~~Shorter might be better.~~ Let's stick with `CredentialProvider` since that's what upstream uses.
- [x] Ability to pass config down from credential provider class, in the case of like boto3 which infers a default region. (implemented but kept undocumented for now)
- [x] Add credential type hints to docs pages

Closes https://github.com/developmentseed/obstore/issues/232, closes #269


kylebarron:
NASA earthdata also has a time limit on its S3 credentials. cc @chuckwondo maybe we can figure out a good example there (although those credentials are only valid for same-region requests IIRC)

chuckwondo:
> NASA earthdata also has a time limit on its S3 credentials. cc @chuckwondo maybe we can figure out a good example there (although those credentials are only valid for same-region requests IIRC)

Sure, let me pull from something I experimented with a while back.

kylebarron:
@chuckwondo in [`4d45b54` (#234)](https://github.com/developmentseed/obstore/pull/234/commits/4d45b5470d1bc78149996743a3a61320e16f7118) I added an example of a custom AWS credential provider, drawing from @abarciauskas-bgse 's code [here](https://github.com/developmentseed/pangeo-forge-staging/blob/6db3af9317594af3adbb72be2d98ef760b3e93f4/recipes/mursst/recipe.py), which I think comes from [here](https://archive.podaac.earthdata.nasa.gov/s3credentialsREADME).

docs/authentication.md
```diff
@@ -101,31 +107,14 @@ class NasaEarthdataCredentialProvider:
...
-        results = requests.get(
-            CREDENTIALS_API, cookies={"accessToken": final.cookies["accessToken"]}
-        )
-        results.raise_for_status()
-
-        creds = results.json()
+        """Request updated credentials."""
+        resp = self.session.get(CREDENTIALS_API, allow_redirects=True, timeout=15)
+        auth_resp = self.session.get(resp.url, allow_redirects=True, timeout=15)
+        creds = auth_resp.json()
```
chuckwondo, line 117: I don't think we should be using this as an example for 2 reasons:

1. The session is never closed
2. It is insecure because you're leaking EDL creds to the CREDENTIALS_API. The EDL creds should be supplied _only_ to the URS (Earthdata Login) URL.

Although your original implementation is more verbose, it is secure, and does not leave dangling resources. (There's no strong motivation for using a session since auth occurs no more than once an hour, or more likely, only once it total for most use cases.)

kylebarron: The original implementation _didn't work_, see https://github.com/developmentseed/obstore/issues/271#issue-2854852119. I'm not sure why it didn't work; the current implementation was suggested by Aimee in #271, and appears to be what the [`earthaccess`](https://github.com/nsidc/earthaccess) Python package is using. 

Using a session seemed to make it easier to automatically manage cookies between the two requests, but it's not a requirement for me. In the async case I add a `close()` method to the credential provider and document that it should be called after all obstore use has finished.

Perhaps this isn't a great use for an example, and we should find something simpler for the docs.

kylebarron:
We just need to update documentation and then we can merge

PR #249, opened 2025-02-07
pyo3-bytes slicing

jessekrubin:
- Adds slicing to `pyo3_bytes::PyBytes`
- Tests for slicing
- Tests for `removeprefix`/`removesuffix`


pyo3-bytes/src/bytes.rs
```diff
@@ -125,20 +186,22 @@ impl PyBytes {
         self.0.as_ref() == other.0.as_ref()
     }
 
-    fn __getitem__(&self, py: Python, key: Bound<PyAny>) -> PyResult<PyObject> {
-        if let Ok(mut index) = key.extract::<isize>() {
-            if index < 0 {
-                index += self.0.len() as isize;
-            }
+    fn __getitem__<'py>(&self, py: Python<'py>, key: BytesGetItemKey<'py>) -> PyResult<PyObject> {
```
kylebarron, line 189: If you're adding the `'py` lifetime anyways, you might as well return `Bound<'py, PyAny>`

pyo3-bytes/src/bytes.rs
```diff
@@ -125,20 +186,22 @@ impl PyBytes {
...
+        match key {
+            BytesGetItemKey::Int(mut index) => {
+                if index < 0 {
+                    index += self.0.len() as isize;
```
kylebarron, line 193: We should probably validate here that the `index` is >= 0 after this addition

pyo3-bytes/src/bytes.rs
```diff
@@ -341,7 +404,7 @@ impl<'py> FromPyObject<'py> for PyBytes {
 ///
 /// This also implements AsRef<[u8]> because that is required for Bytes::from_owner
 #[derive(Debug)]
-struct PyBytesWrapper(Option<PyBuffer<u8>>);
+pub struct PyBytesWrapper(Option<PyBuffer<u8>>);
```
kylebarron, line 407: Why should this be pub?

pyo3-bytes/src/bytes.rs
```diff
@@ -424,3 +487,12 @@ impl std::fmt::Debug for PyBytes {
         Ok(())
     }
 }
+
+/// A key for the `__getitem__` method of `PyBytes` - int/slice
+#[derive(FromPyObject)]
+pub enum BytesGetItemKey<'py> {
```
kylebarron, line 493: Why should this be `pub`?

tests/test_bytes.py
```diff
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
+import typing as t
```
kylebarron, line 3: Since the only import you use is `Generator`, just use `from typing import Generator`

tests/test_bytes.py
```diff
@@ -25,3 +29,53 @@ def test_uno_byte_bytes_repr(b: bytes) -> None:
...
+        rs_bytes = Bytes(py_bytes)
+        for start, stop, step, _sliced in _bytes_slices(py_bytes):
+            new_py = py_bytes[start:stop:step]
+            new_rs = rs_bytes[start:stop:step]
+            assert new_rs == new_py
+
+
+def _bytes_slices(
+    b: bytes, range_buffer: int = 3
+) -> t.Generator[tuple[int, int, int, bytes], None, None]:
```
kylebarron, line 71: Also I think this typing can be simplified to just `-> Iterable[tuple[]]` instead of `Generator` IIRC?

pyo3-bytes/bytes.pyi
```diff
@@ -25,7 +26,10 @@ class Bytes(Buffer):
     def __buffer__(self, flags: int) -> memoryview[int]: ...
     def __contains__(self, other: Buffer) -> bool: ...
     def __eq__(self, other: object) -> bool: ...
-    def __getitem__(self, other: int) -> int: ...
+    @overload
+    def __getitem__(self, key: int, /) -> int: ...
+    @overload
+    def __getitem__(self, key: slice, /) -> Bytes: ...
```
kylebarron, line 32: You can't have only overloads; you need the overall `__getitem__` too IIRC

jessekrubin: not in `.pyi` files

tests/test_bytes.py
jessekrubin: totes

tests/test_bytes.py
jessekrubin: ack. its a habit for me.

pyo3-bytes/src/bytes.rs
jessekrubin: vestigial from experimenting around.

pyo3-bytes/src/bytes.rs
jessekrubin: whoops. This is a remnant of me trying to figure out how to make taking a `pyo3_bytes::PyBytes` or `pyo3::...::PyBytes` object when you only need `&[u8]` -- I was going to write up my findings and send them to you -- basically taking a `PyBytes` when you only need a ref to the slice has a large perf hit.

tests/test_bytes.py
jessekrubin: boom fixed

pyo3-bytes/src/bytes.rs
kylebarron: > basically taking a PyBytes when you only need a ref to the slice has a large perf hit

hmm, can you expand on this?

pyo3-bytes/bytes.pyi
kylebarron: I tried this and it broke mkdocstrings; so I think we need it if it's going into the documentation.

If you add a short docstring here, then it should get rendered in mkdocs, and then it should break if I'm correct.

pyo3-bytes/src/bytes.rs
jessekrubin: BABOOM
https://github.com/developmentseed/obstore/issues/251

pyo3-bytes/bytes.pyi
jessekrubin: oh no. I _can_ add one, but adding it breaks mypy:

```txt
jesse@corbett[[pyo3-bytes]]$ mypy bytes.pyi
bytes.pyi:33: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
Found 1 error in 1 file (checked 1 source file)
```

kylebarron: You don't provide an implementation here...

It would be
```py
    def __getitem__(self, key: int | slice, /) -> int | Bytes: ...
```
There are various other implementations of `overload` already. Does mypy fail on them?

jessekrubin: ack. I added an implementation before running mypy... I assume it does

```diff
jesse@corbett[[pyo3-bytes]]$ git diff
diff --git a/pyo3-bytes/bytes.pyi b/pyo3-bytes/bytes.pyi
index c02e0e3..468f101 100644
--- a/pyo3-bytes/bytes.pyi
+++ b/pyo3-bytes/bytes.pyi
@@ -30,6 +30,7 @@ class Bytes(Buffer):
     def __getitem__(self, key: int, /) -> int: ...
     @overload
     def __getitem__(self, key: slice, /) -> Bytes: ...
+    def __getitem__(self, key: int | slice, /) -> int | Bytes: ...
     def __mul__(self, other: Buffer) -> int: ...
     def __len__(self) -> int: ...
     def __repr__(self) -> str: ...
```

err

```txt
jesse@corbett[[pyo3-bytes]]$ mypy bytes.pyi
bytes.pyi:33: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
Found 1 error in 1 file (checked 1 source file)
```

jessekrubin: >  Does mypy fail on them?

Looks like it does:

```
jesse@corbett[[obstore]]$ mypy python/
python/obstore/_bytes.pyi:33: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
python/obstore/store/_aws.pyi:3: error: Cannot find implementation or library stub for module named "boto3"  [import-not-found]
python/obstore/store/_aws.pyi:4: error: Cannot find implementation or library stub for module named "boto3.session"  [import-not-found]
python/obstore/store/_aws.pyi:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
python/obstore/store/_aws.pyi:5: error: Cannot find implementation or library stub for module named "botocore"  [import-not-found]
python/obstore/store/_aws.pyi:6: error: Cannot find implementation or library stub for module named "botocore.session"  [import-not-found]
python/obstore/store/__init__.pyi:56: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
python/obstore/_sign.pyi:25: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
python/obstore/_sign.pyi:50: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
python/obstore/_sign.pyi:63: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
python/obstore/_list.pyi:4: error: Cannot find implementation or library stub for module named "arro3.core"  [import-not-found]
python/obstore/_list.pyi:99: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
python/obstore/_list.pyi:211: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
python/obstore/_list.pyi:251: error: An implementation for an overloaded function is not allowed in a stub file  [misc]
python/obstore/fsspec.py:29: error: Cannot find implementation or library stub for module named "fsspec.asyn"  [import-not-found]
python/obstore/fsspec.py:29: error: Cannot find implementation or library stub for module named "fsspec"  [import-not-found]
python/obstore/fsspec.py:30: error: Cannot find implementation or library stub for module named "fsspec.spec"  [import-not-found]
python/obstore/fsspec.py:126: error: Argument 1 to "append" of "list" has incompatible type "Coroutine[Any, Any, list[Bytes]]"; expected "Coroutine[Any, Any, list[bytes]]"  [arg-type]
python/obstore/fsspec.py:135: error: "bytes" has no attribute "to_bytes"  [attr-defined]
python/obstore/fsspec.py:198: error: Cannot determine type of "loc"  [has-type]
Found 19 errors in 6 files (checked 22 source files)
```

kylebarron: It looks like mypy fails on all the pyi files in this repo 🤷‍♂️  (I've been using pylance in vscode)

kylebarron: If you'd like to dig into how to ensure mkdocs builds even without that last non-overloaded definition, be my guest

jessekrubin: oh well!

I could if I find some spare time soon, but as marie kondo has stated "mypy and python-type-annotations do not bring me joy".

jessekrubin:
@kylebarron Fixed!

tests/test_bytes.py
```diff
@@ -21,7 +30,66 @@ def test_repr():
...
+    def test_slice_o_bytes(self, py_bytes: bytes) -> None:
+        """Run slicing on both bytes and Bytes and assert they are equal."""
+        rs_bytes = Bytes(py_bytes)
+        for start, stop, step, _sliced in _bytes_slices(py_bytes):
+            new_py = py_bytes[start:stop:step]
+            new_rs = rs_bytes[start:stop:step]
+            assert new_rs == new_py
+
+
+def _bytes_slices(
```
kylebarron, line 82: If this is only used in `TestBytesSlice` and isn't intended to be used by other tests in the future, can you indent this as part of the class and decorate it with `@staticmethod`?

pyo3-bytes/src/bytes.rs
```diff
@@ -62,6 +63,66 @@ impl PyBytes {
...
+
+    /// Slice the underlying buffer using a Python slice object
+    ///
+    /// This should behave the same as Python's byte slicing:
+    ///     - `ValueError` if step is zero
+    ///     - Negative indices a-ok
+    ///     - If start/stop are out of bounds, they are clipped to the bounds of the buffer
+    ///     - If start > stop, the slice is empty
+    ///
+    /// This is NOT exposed to Python under the `#[pymethods]` impl
```
kylebarron, line 75:
```suggestion
    /// This is a helper method used by `__getitem__` and is not directly exposed to Python
```

pyo3-bytes/src/bytes.rs
```diff
@@ -62,6 +63,66 @@ impl PyBytes {
...
+    fn slice(&self, slice: &Bound<'_, PySlice>) -> PyResult<PyBytes> {
+        let len_isize = self.0.len() as isize;
+        let psi = slice.indices(len_isize)?;
```
kylebarron, line 78: What does `psi` stand for? Could we maybe have a more descriptive name?

kylebarron: Or, it looks like `psi` is never used apart from the next line, so it might be clearer to create a new scope
```rs
let (start, stop, step) = { 
    let indices = slice.indices(len_isize)?;
    (psi.start, psi.stop, psi.step)
};
```

pyo3-bytes/src/bytes.rs
```diff
@@ -62,6 +63,66 @@ impl PyBytes {
...
+        let (start, stop, step) = (psi.start, psi.stop, psi.step);
+        if step == 0 {
+            return Err(PyValueError::new_err("step is zero"));
+        }
+
+        // I think this is right!?
+        let new_cap_usize = if (step > 0 && stop > start) || (step < 0 && stop < start) {
```
kylebarron, line 85: What is `new_cap_usize`?

pyo3-bytes/src/bytes.rs
```diff
@@ -62,6 +63,66 @@ impl PyBytes {
...
+    /// This is NOT exposed to Python under the `#[pymethods]` impl
+    fn slice(&self, slice: &Bound<'_, PySlice>) -> PyResult<PyBytes> {
+        let len_isize = self.0.len() as isize;
+        let psi = slice.indices(len_isize)?;
+        let (start, stop, step) = (psi.start, psi.stop, psi.step);
+        if step == 0 {
+            return Err(PyValueError::new_err("step is zero"));
+        }
+
+        // I think this is right!?
```
kylebarron, line 84: Can we have a test to test this case specifically? I don't know what it's doing

pyo3-bytes/src/bytes.rs
kylebarron: I assume `cap` means `capacity`? There's no reason for `usize` to be part of the variable name. Just spell it out with `new_capacity`

pyo3-bytes/src/bytes.rs
```diff
@@ -62,6 +63,66 @@ impl PyBytes {
...
+    ///
+    /// This should behave the same as Python's byte slicing:
+    ///     - `ValueError` if step is zero
+    ///     - Negative indices a-ok
+    ///     - If start/stop are out of bounds, they are clipped to the bounds of the buffer
+    ///     - If start > stop, the slice is empty
+    ///
+    /// This is NOT exposed to Python under the `#[pymethods]` impl
+    fn slice(&self, slice: &Bound<'_, PySlice>) -> PyResult<PyBytes> {
+        let len_isize = self.0.len() as isize;
```
kylebarron, line 77: Similarly, we can see the types with rust-analyzer, and we don't need to have the type be part of the variable name. So just name this like `bytes_length`.

pyo3-bytes/src/bytes.rs
```diff
@@ -62,6 +63,66 @@ impl PyBytes {
...
+            let out = self.0.slice(..);
+            let py_bytes = PyBytes(out);
+            return Ok(py_bytes);
+        }
+
+        if step == 1 && start >= 0 && stop <= len_isize && start < stop {
+            let out = self.0.slice(start as usize..stop as usize);
+            let py_bytes = PyBytes(out);
+            return Ok(py_bytes);
+        }
```
kylebarron, line 106: I think it would be clearer to combine these cases of `step` into a single if/else case block

```rs
if step == 0 {

} else if step == 1 {
 .. above code
} else if step > 0 {
let mut new_buf = BytesMut::with_capacity(new_cap_usize);
...
} else if step < 0 {
...
} else {
	unreachable!("Value of step: {}", step)
}
```

It's fine to always compute the new capacity before checking if `step == 0`, because we assume that's an error if it is passed in.

pyo3-bytes/src/bytes.rs
jessekrubin: Which one? step=0? I think there is a test for that.

pyo3-bytes/src/bytes.rs
jessekrubin: True that. Old habit!

pyo3-bytes/src/bytes.rs
jessekrubin: @kylebarron fixed this... I dont think we actually have to check for step==0, as the `.indices` call on pyslice throws that error for us.

kylebarron:
Looks good except for the type hint:
```
Argument of type "slice[int, int, int]" cannot be assigned to parameter "other" of type "int" in function "__getitem__"
  "slice[int, int, int]" is not assignable to "int"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

kylebarron:
Looks like pylance can't handle only overloads 🤷‍♂️

kylebarron (approved):
LGTM; thanks!

jessekrubin:
> Looks good except for the type hint:
> 
> ```
> Argument of type "slice[int, int, int]" cannot be assigned to parameter "other" of type "int" in function "__getitem__"
>   "slice[int, int, int]" is not assignable to "int"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
> ```

I might have changed that when I was a bit fried at the end of the day... so herm. 

> Looks like pylance can't handle only overloads 🤷‍♂️

oy vey, what can you do! 

Where can I pick your brain about pyo3 related things? discussions on this repo? email? TY

kylebarron:
Maybe start with a discussion thread here

PR #272, opened 2025-02-15
Print full error message for GenericError

### LINKED ISSUE #169: Improve error printing
I was talking with @vincentsarago and he mentioned he got an error that looked like:

```
GenericError: Generic {
    store: "S3",
    source: BareRedirect,
}
```

Looking at the upstream implementation [I see](https://github.com/apache/arrow-rs/blob/6aaff7e38a573f797b31f89f869c3706cbe26e37/object_store/src/client/retry.rs#L29C1-L35C1)
 
```rs
/// Retry request error
#[derive(Debug, thiserror::Error)]
pub enum Error {
    #[error("Received redirect without LOCATION, this normally indicates an incorrectly configured region")]
    BareRedirect,
```

Perhaps we should update our errors to use the `to_string` impl? Note that this was previously changed in https://github.com/developmentseed/obstore/pull/54 https://github.com/developmentseed/obstore/blame/78860c6e71255bdf06c973b9b9435cdedc2e3fdd/pyo3-object_store/src/error.rs#L112

weiji14:
## What I am changing

- Make BareRedirect error message more informative

## How I did it

- Change `format!("{err:#?}")` to `err.to_string()`

## How you can test it

- Run `uv run maturin develop -m obstore/Cargo.toml`, and then try opening some file object that requires setting AWS_REGION

## Related Issues

- Resolves #169

### DIFF
--- a/pyo3-object_store/src/error.rs
+++ b/pyo3-object_store/src/error.rs
@@ -111,7 +111,7 @@ impl From<PyObjectStoreError> for PyErr {
                 object_store::Error::Generic {
                     store: _,
                     source: _,
-                } => GenericError::new_err(format!("{err:#?}")),
+                } => GenericError::new_err(err.to_string()),
                 object_store::Error::NotFound { path: _, source: _ } => {
                     PyFileNotFoundError::new_err(format!("{err:#?}"))
                 }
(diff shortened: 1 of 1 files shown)


weiji14:
Let me know if you'd like some unit tests for this (and where to add them).

kylebarron:
This seems fine. I'm not sure if we need tests for this, or, if we do, what the best way to have tests is.

PR #287, opened 2025-02-26
Remove `S3Store._from_native` (Revert #203)

kylebarron:
### Change list

- Remove `S3Store._from_native`
- Remove `pickle_safe` flag, since now `S3Store` is always pickle safe, as long as the Python credential provider is pickle-safe.

### Background

In version 0.4 in https://github.com/developmentseed/obstore/pull/203 we added `S3Store._from_native`. This statically links to the `aws-config` crate

It caused the size of the wheel to go from 3.4MB (8.22MB uncompressed) to 4.4MB (11.3MB uncompressed) on macOS. That's not huge on its own, but natively linking to GCS and Azure crates in the future would add additional bloat.

More importantly, however, `_from_native` expanded the scope of the library. Now that we support Python-based credential-providers (#234), we no longer need to manage all sorts of custom credential handling ourselves in Rust. 

We'll continue to support any credential handling that `object_store` does natively, but we won't add any additional credential handling on the rust side ourselves.

In practice, removing `_from_native` mainly means that disk-based credential handling no longer happens on the Rust side.

cc @ion-elgreco


ion-elgreco:
Yeah this makes more sense! Credential vending should be lightweight anyways, so not much penalty to keep it in python

PR #320, opened 2025-03-05
WIP: vendor metrics code

kylebarron:
For #122 

This is pretty involved. Leaving this for some future date.

It would really be preferable if this were some external crate


kylebarron:
Pending #339 perhaps this sort of middleware could even be done on the Python side.

PR #330, opened 2025-03-07
Obspec interface specification

kylebarron:
Thoughts

- Move to standalone repo https://github.com/developmentseed/obspec
- Make `ObjectStore` and `AsyncObjectStore` type hints as the union of all sync/async protocols.


kylebarron:
Moved to https://github.com/developmentseed/obspec

Issue #160, opened 2025-01-20
Document design decision for functional API


ion-elgreco:
It felt a bit unintuitive to have to do provide the store each time to execute a method:

```py
import obstore as obs
from obstore.store import AzureStore
store = AzureStore()
obs.put(store, ....)
```

Instead of having the methods on the store directly akin to the rust implementation:

```py
import obstore as obs
from obstore.store import AzureStore
store = AzureStore()
store.put(....)
```

kylebarron:
Yes, I agree that it would be _nice_ to have the methods directly on the class, however this is a very intentional design decision,

### Simpler Rust code

On the Rust side, each Python class is a separate `struct`. A Pyclass can't implement a trait, so the only way to implement the same methods on multiple Rust structs without copy-pasting is by having a macro. That isn't out of the question, however it does hamper extensibility, and having one and only one way to call commands is simpler to maintain.

### Simpler Middlewares

In https://github.com/developmentseed/obstore/pull/117 we added a binding for `PrefixStore`. Because we use object store classes functionally, we only needed 20 lines of Rust code:
https://github.com/developmentseed/obstore/blob/b40d59b4e060ba4fd3dc69468b3ba7da1149758e/pyo3-object_store/src/prefix.rs#L10-L25

If we exposed methods on an `S3Store`, then those methods would be lost whenever you apply a middleware around it, such as `PrefixStore(S3Store(...))`. So we'd have to ensure those same methods are also installed onto every middleware or other wrapper.

### External FFI for ObjectStore

There was recently [discussion on Discord](https://discord.com/channels/885562378132000778/885562378132000781/1328392836353360007) about the merits of having a stable FFI for `ObjectStore`. If this comes to fruition in the future, then by having a functional API we could seamlessly use _third party_ ObjectStore implementations or middlewares, with no Python overhead.

I use a similar functional API in other Python bindings, especially in cases with zero-copy FFI, such as https://kylebarron.dev/geo-index/latest/api/rtree/#geoindex_rs.rtree.search (where the spatial index is passed in as the first argument instead) and https://kylebarron.dev/arro3/latest/api/compute/#arro3.compute.cast where the `cast` is not a method on the Arrow Array.

### Smaller core for third-party Rust bindings

This repo has twin goals:

1. Provide bindings to `object_store` for _Python users_ who want a _Python API_.
[...]

Issue #247, opened 2025-02-07
Add CachingStore


ion-elgreco:
Was talking with @danielgafni on this. Would be nice if we can have a CachingStore, would work similarly to LimitStore, where we can wrap the inner store.

The CachingStore should be configurable with cache_size, tmp location and so forth.

I know there are a couple example implementations on this:

- Caching layer for immutable commit files https://github.com/delta-io/delta-rs/blob/5d2cf488ff7f69393b1e3c73bf72f1dcdd832768/crates/core/src/kernel/snapshot_next/cache.rs
- Bytes aware caching layer (focused on parquet reading): https://github.com/splitgraph/seafowl/blob/main/src/object_store/cache.rs

So perhaps we can add a general-purpose caching layer for immutable objects, and a version for potential mutable objects and a version for parquet specifically.

kylebarron:
I agree the caching would be nice, and would be a logical extension of the existing functionality.

But caching gets really complex really quickly, and I don't want the full responsibility in this repo of managing that caching. Ideally I'd prefer that to be shipped as a standalone Python library, but given that `object_store` doesn't have stable FFI, I don't think that's really possible.

So my suggestion would be to create a third party Rust repo (`object-store-cache`?), which means that that caching middleware could be used by obstore but also by any other Rust library, and then once that repo is stable, we can expose a wrapper through obstore.

danielgafni:
I think your proposal makes sense. 
I will think if I can manage this alone (my experience with Rust is pretty limited, mostly as a reader) - creating a standalone project is more complicated than contributing to an existing one. 
Do any of you guys have any interest in working on this together?

ion-elgreco:
I am a bit swamped in my weekends for the next 1-2 months. After that I should be able to support in dev time. In the mean time I can help with reviews though

danielgafni:
Alright. I started working on something and will ask you guys to take a look soon :)

kylebarron:
Ref https://github.com/apache/arrow-rs/issues/7251

kylebarron:
[Obspec](https://github.com/developmentseed/obspec) is ready to go for an 0.1 release; it's only waiting on me to write a release post (the "why") accompanying an 0.1 release. I'm going to close this as I want obstore itself to remain small and tied to the object_store implementation.

Perhaps when https://github.com/apache/arrow-rs-object-store/issues/14 is implemented and there's a stable implementation of caching utilities we can reconsider this.

PR #331, opened 2025-03-07
Add class methods wrapper

kylebarron:
This PR allows all functional methods, like `get`, `head`, `list`, to also be called as methods on each store class.

I want obstore to define an "object storage interface spec" that allows applications to depend on an obstore-like interface, but which allows middlewares in between. So for example, metrics (https://github.com/developmentseed/obstore/issues/105) or caching (https://github.com/developmentseed/obstore/issues/247) could be implemented externally **on the Python side**. And a library depending on an obstore like interface wouldn't need to know anything about middlewares, as long as the input conforms to the correct protocol. 

I started exploring these object storage interface spec ideas in https://github.com/developmentseed/obstore/pull/330. I think the API I'm considering is just a collection of protocols. E.g.
```py
class Head(Protocol):
    def head(self) -> HeadResult: ...

class HeadAsync(Protocol):
    async def head(self) -> HeadResult: ...
```
etc. Then a downstream application could pick and choose which APIs they require by doing
```
class RequiredObspecMethods(Head, Get, List, Protocol):
    pass
```
which would union the required methods.

But this obspec idea requires a class-based approach, with named functions that can match the given protocol methods.

Hence, this requires revisiting #160, to consider exposing methods on the Python based classes. I think revisiting this is worthwhile.

### Change list

- Allow subclassing of raw Rust-exported python classes, like `S3Store`
- Ensure that the raw functional API imports from `_store`, so that the functional API supports both the non subclassed and subclassed variants.
- Convert the store-local `from_url`, i.e. `S3Store.from_url` to respect subclasses. That is, use `cls()` to construct the class instance.
    - Also change the typing to return `-> typing.Self` instead of the concrete class itself.
- Add a top-level `from_url` method in Python, so that `obstore.store.from_url` can construct the _subclassed_ versions of each class.
    - Add a private `parse_scheme` helper function exported from Rust to make sure we parse the URLs in the same way that `object_store` does.
- Add class methods to docs
- Update the `register_store_module` function so that the user of `pyo3-object_store` can decide whether to register `.store` or `._store`
- Update tests to use this new class-based API

----

There are a bunch of moving pieces here:

- We define a mixin, `_ObjectStoreMixin`, which defines the method wrappers. So these methods are defined _in terms_ of the pure functions.
- Because these mixins are defined on the Python side, we don't increase the API surface exported to other Rust libraries (via pyo3-object_store).
- We **subclass** the underlying Rust classes **instead of wrapping**. This means that these same classes can be passed in to the existing functional API (so it's backwards compatible). 
- Pickling still works because the `__module__` attribute points to the _top-level Python class_, not the core Rust class (does cloudpickle work too?, hopefully)

Closes https://github.com/developmentseed/obstore/issues/160

cc @ion-elgreco


kylebarron:
I'm not sure why, but the exported classes don't seem to be the correct (subclassed) ones

Issue #267, opened 2025-02-14
Automatically inherit Azure credentials from environment


daviewales:
With Polars, which also uses the `object_store` crate, I can authenticate to Azure automatically, based on credentials found in my environment.
For example, if a client secret environment variable exists, it will use that. If I am authenticated with Azure CLI, it will use that.
This means that my code looks like this, and works both for local testing and on the server with environment credentials:

``` python
import polars as pl
pl.read_parquet('abfs://container@account.dfs.core.windows.net/path/to/file.parquet')
```

This is handy, because it means that I can write code which is completely agnostic to the location of the file.
The code above works just as well if I swap out the filepath for a local file, because I didn't need to specify any Azure-specific parameters.

`obstore` is getting much closer to this ideal.
However, for Azure at least, I still need to give it a hint to detect my Azure CLI credentials:

``` python
import obstore
from obstore.store import from_url
store = from_url('abfs://container@account.dfs.core.windows.net/path/to/directory', azure_use_azure_cli=True)
obstore.list(store).collect()
```

Ideally, I would be able to do this instead:

``` python
import obstore
from obstore.store import from_url
store = from_url('abfs://container@account.dfs.core.windows.net/path/to/directory')
obstore.list(store).collect()
```

Note that Polars uses [azure.identity.DefaultAzureCredential](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python) as the [default credential provider](https://github.com/pola-rs/polars/blob/da1b2fc889f28769dd722e55b8f84c04f64dff0c/py-polars/polars/io/cloud/credential_provider/_providers.py#L159) for Azure, as it automatically finds available credentials. This ensures consistent credential resolution across tools.

(It's not a hard dependency on `azure.identity`. An import error is [only raised](https://github.com/pola-rs/polars/blob/da1b2fc889f28769dd722e55b8f84c04f64dff0c/py-polars/polars/io/cloud/credential_provider/_providers.py#L290) if you try to use Azure credential functionality and don't have `azure.identity` installed.)

kylebarron:
You should be able to set `AZURE_USE_AZURE_CLI=True` as an environment variable.

daviewales:
[...]

PR #343, opened 2025-03-12
Add Azure credential providers

daviewales:
## What I am changing

- Add Azure credential providers

## How I did it

- Implement sync provider with `azure.identity` and async provider with `azure.identity.aio`
- Note: I used `credential` rather than `credentials` for the `__init__` argument, as this matches other Azure libraries. Let me know if I should change this to `credentials` to match the existing Google credential provider.
- I added simple token caching based on the token expiry time. If the token does not exist or expires in less than 5 minutes, fetch a new token, otherwise reuse the cached one.

## How you can test it

- Follow the examples in the docstrings.

## Related Issues

- Improves #267.

## WIP

This PR is WIP because although the credential_provider classes _appear_ to work when testing them directly, I get the following error I try to list items after initialising a store with them:

``` python
import obstore
from obstore.auth.azure import AzureAuthCredentialProvider
credential_provider = AzureAuthCredentialProvider()
store = obstore.from_url("abfs://test@example.dfs.core.windows.net", credential_provider=credential_provider)
obstore.list(store).collect()

# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# obstore.exceptions.GenericError: Generic External Azure credential provider error: RuntimeError: no running event loop
# 
# Debug source:
# Generic {
#     store: "External Azure credential provider",
#     source: PyErr {
#        type: <class 'RuntimeError'>,
#         value: RuntimeError('no running event loop'),
#         traceback: None,
#     },
# }
```

This sounds like an async error, but I'm not sure why, because I'm not using the async credential provider here.
Also, even if I do use the async credential provider, I get the same error.
This error appears to be raised from: `pyo3-object_store\src\azure\credentials.rs:227`
Is there a trick to enable tracebacks with line numbers, etc?
Can you shed some light on this?

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -89,7 +89,37 @@ store = GCSStore("bucket_name", credential_provider=credential_provider)
 
 Refer to [`obstore.auth.google`](api/auth/google.md).
 
-<!-- #### Using `azure.identity` -->
+#### `azure.identity`
+
+You can use the [`AzureCredentialProvider`][obstore.auth.azure.AzureCredentialProvider] to use [`azure.identity`][] to handle credentials.
+
+``` python
+import obstore
+from obstore.auth.azure import AzureCredentialProvider
+from obstore.store import AzureStore
+
+credential_provider = AzureAsyncCredentialProvider(credential=...)
+store = AzureStore("container", credential_provider=credential_provider)
+print(obstore.list(store).collect())
+```
+
+Alternatively, you can use [`AzureAsyncCredentialProvider`][obstore.auth.azure.AzureAsyncCredentialProvider] with the async API:
+
+``` python
+import asyncio
+import obstore
+from obstore.auth.azure import AzureCredentialProvider
+from obstore.store import AzureStore
+
+credential_provider = AzureAsyncCredentialProvider(credential=...)
+store = AzureStore("container", credential_provider=credential_provider)
+
+async def fetch_blobs():
+    blobs = await obstore.list(store).collect_async()
+    print(blobs)
+
+asyncio.run(fetch_blobs())
+```
 
 ### Custom Authentication
 
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -77,13 +77,13 @@ Refer to [`obstore.auth.boto3`](api/auth/boto3.md).
 
 #### google.auth
 
-You can use the [`GoogleAuthCredentialProvider`][obstore.auth.google.GoogleAuthCredentialProvider] to use [`google.auth`][] to handle credentials.
+You can use the [`GoogleCredentialProvider`][obstore.auth.google.GoogleCredentialProvider] to use [`google.auth`][] to handle credentials.
 
 ```py
-from obstore.auth.google import GoogleAuthCredentialProvider
+from obstore.auth.google import GoogleCredentialProvider
 from obstore.store import GCSStore
 
-credential_provider = GoogleAuthCredentialProvider(credentials=...)
+credential_provider = GoogleCredentialProvider(credentials=...)
 store = GCSStore("bucket_name", credential_provider=credential_provider)
 ```
 
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -119,6 +119,8 @@ async def fetch_blobs():
     print(blobs)
 
 asyncio.run(fetch_blobs())
+
+Refer to [`obstore.auth.azure`](api/auth/azure.md).
 ```
 
 ### Custom Authentication
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -94,20 +94,20 @@ Refer to [`obstore.auth.google`](api/auth/google.md).
 You can use the [`AzureCredentialProvider`][obstore.auth.azure.AzureCredentialProvider] to use [`azure.identity`][] to handle credentials.
 
 ``` python
-import obstore
+import obstore as obs
 from obstore.auth.azure import AzureCredentialProvider
 from obstore.store import AzureStore
 
 credential_provider = AzureAsyncCredentialProvider(credential=...)
 store = AzureStore("container", credential_provider=credential_provider)
-print(obstore.list(store).collect())
+print(obs.list(store).collect())
 ```
 
 Alternatively, you can use [`AzureAsyncCredentialProvider`][obstore.auth.azure.AzureAsyncCredentialProvider] with the async API:
 
 ``` python
 import asyncio
-import obstore
+import obstore as obs
 from obstore.auth.azure import AzureCredentialProvider
 from obstore.store import AzureStore
 
@@ -115,7 +115,7 @@ credential_provider = AzureAsyncCredentialProvider(credential=...)
 store = AzureStore("container", credential_provider=credential_provider)
 
 async def fetch_blobs():
-    blobs = await obstore.list(store).collect_async()
+    blobs = await obs.list(store).collect_async()
     print(blobs)
 
 asyncio.run(fetch_blobs())
--- a/obstore/python/obstore/auth/azure.py
+++ b/obstore/python/obstore/auth/azure.py
@@ -5,7 +5,7 @@
 
 from __future__ import annotations
 
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, timedelta
 from typing import TYPE_CHECKING
 
 import azure.identity
@@ -57,8 +57,7 @@ if TYPE_CHECKING:
 
 
 class AzureAuthCredentialProvider:
-    """A CredentialProvider for [AzureStore][obstore.store.AzureStore] that uses
-    [`azure.identity`][].
+    """A CredentialProvider for [AzureStore][obstore.store.AzureStore] that uses [`azure.identity`][].
 
     This credential provider uses `azure-identity`, and will error if this cannot
     be imported.
@@ -74,7 +73,7 @@ class AzureAuthCredentialProvider:
     ```
 
     [`azure.identity`]: https://learn.microsoft.com/en-us/python/api/overview/azure/identity-readme
-    """
+    """  # noqa: E501
 
     credential: AzureCredentialUnionType
 
@@ -90,10 +89,14 @@ class AzureAuthCredentialProvider:
             credential: Credential to use for this provider. Defaults to `None`,
                 in which case [`azure.identity.DefaultAzureCredential`][] will be
                 called to find default credentials.
+            scopes: Scopes required by the access token. If not specified,
+                ["https://storage.azure.com/.default"] will be used by default.
+            tenant_id: Optionally specify the Azure Tenant ID which will be passed to
+                the credential's `get_token` method.
 
         [`azure.identity.DefaultAzureCredential`]: https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential
-        """
 
+        """
         self.credential = credential or azure.identity.DefaultAzureCredential()
 
         # Use the Azure Storage scope by default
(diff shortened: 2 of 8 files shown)


obstore/python/obstore/auth/azure.py
```diff
@@ -0,0 +1,200 @@
...
+        return {
+            "token": self.token.token,
+            "expires_at": datetime.fromtimestamp(self.token.expires_on, UTC),
+        }
+
+
+class AzureAuthAsyncCredentialProvider:
+    """An async CredentialProvider for [AzureStore][obstore.store.AzureStore] that uses [`azure.identity`][].
+
+    This credential provider uses `azure-identity` and `aiohttp`, and will error if
```
kylebarron, line 134: Is `aiohttp` used under the hood?

obstore/python/obstore/auth/azure.py
```diff
@@ -0,0 +1,200 @@
...
+        | azure.identity.aio.EnvironmentCredential
+        | azure.identity.aio.ManagedIdentityCredential
+        | azure.identity.aio.OnBehalfOfCredential
+        | azure.identity.aio.SharedTokenCacheCredential
+        | azure.identity.aio.VisualStudioCodeCredential
+        | azure.identity.aio.WorkloadIdentityCredential
+    )
+
+
+class AzureAuthCredentialProvider:
```
kylebarron, line 59: I named it `GoogleAuthCredentialProvider` because the module is `google.auth`. So this would be `AzureIdentityCredentialProvider` because the module used is `azure.identity`. That said, that name is super long so I'm open to alternate suggestions.

obstore/python/obstore/auth/azure.py
```diff
@@ -0,0 +1,200 @@
...
+            tenant_id: Optionally specify the Azure Tenant ID which will be passed to
+                the credential's `get_token` method.
+
+        [`azure.identity.DefaultAzureCredential`]: https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential
+
+        """
+        self.credential = credential or azure.identity.DefaultAzureCredential()
+
+        # Use the Azure Storage scope by default
+        self.scopes = scopes or ["https://storage.azure.com/.default"]
```
kylebarron, line 103: Can you put this at the top level as
```py
DEFAULT_SCOPES = ("https://storage.azure.com/.default", )
```
(tuple, not list, to be an immutable sequence)

obstore/python/obstore/auth/azure.py
```diff
@@ -0,0 +1,200 @@
...
+
+    [`azure.identity`]: https://learn.microsoft.com/en-us/python/api/overview/azure/identity-readme
+    """  # noqa: E501
+
+    credential: AzureCredentialUnionType
+
+    def __init__(
+        self,
+        credential: AzureCredentialUnionType | None = None,
+        scopes: list[str] | None = None,
```
kylebarron, line 83: And then here, this can be
```suggestion
        scopes: Iterable[str] = DEFAULT_SCOPES,
```

kylebarron:
Thanks for the PR!

> * I added simple token caching based on the token expiry time. If the token does not exist or expires in less than 5 minutes, fetch a new token, otherwise reuse the cached one.

You don't need to do this! This is automatically done on the Rust side. 

https://github.com/developmentseed/obstore/blob/b4e8d97cbfad18a6d3b4d7ad7e72f60af4b3f17f/pyo3-object_store/src/credentials.rs#L6-L26

> * Note: I used `credential` rather than `credentials` for the `__init__` argument, as this matches other Azure libraries. Let me know if I should change this to `credentials` to match the existing Google credential provider.

I think whatever is most similar to the native library is best.

> This PR is WIP because although the credential_provider classes _appear_ to work when testing them directly, I get the following error I try to list items after initialising a store with them:

Hmm thanks, I'm not sure at a glance what to make of this. 

I was wondering if maybe my manual testing had worked because I was using ipython instead of python, but the Boto3CredentialProvider worked for me in the normal `Python` shell as well.

That does look like an async error, but I'm not exactly sure why it's happening.

I think the most likely error is that it's failing to convert the result here: https://github.com/developmentseed/obstore/blob/85bfb3a8c41a314904afe8a6fa619365be993116/pyo3-object_store/src/azure/credentials.rs#L203-L205
so it's probably assuming that your result is an awaitable, and then awaiting on it fails.

We should probably figure out a way to test that data conversion directly

kylebarron:
> I think the most likely error is that it's failing to convert the result here:

Indeed, this happens when there's no event loop running and it tries to await the result of a non-awaitable:
```
>>> def credential_provider():
...     return "test"
...
>>> store = S3Store(
...     "ds-wheels",
...     credential_provider=credential_provider,
...     region="us-east-1",
... )
>>> obs.list(store).collect()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
obstore.exceptions.GenericError: Generic External AWS credential provider error: RuntimeError: no running event loop

Debug source:
Generic {
    store: "External AWS credential provider",
    source: PyErr {
        type: <class 'RuntimeError'>,
        value: RuntimeError('no running event loop'),
        traceback: None,
    },
}
```

When I run that code in IPython, the terminal hangs forever, because it's trying to await the result of a non-awaitable. Let me add better error handling

obstore/python/obstore/auth/azure.py
daviewales: Yes it is.

obstore/python/obstore/auth/azure.py
daviewales: How do you feel about `AzureCredentialProvider`?

kylebarron: That's fine; maybe we should rename the `GoogleAuth` too

daviewales: Done. (For Google too.)

obstore/python/obstore/auth/azure.py
daviewales: Done.

obstore/python/obstore/auth/azure.py
daviewales: Done.

daviewales:
On latest main, the sync credential provider now works!

``` python
import obstore
from obstore.auth.azure import AzureCredentialProvider
credential_provider = AzureCredentialProvider()
store = obstore.store.from_url("abfs://test@example.dfs.core.windows.net", credential_provider=credential_provider)
obstore.list(store).collect()

# [{'path': 'example/directory/example.parquet', ...
```

The async credential provider fails with this error message:

``` python
import obstore
from obstore.auth.azure import AzureAsyncCredentialProvider
credential_provider = AzureAsyncCredentialProvider()
store = obstore.store.from_url("abfs://test@example.dfs.core.windows.net", credential_provider=credential_provider)
obstore.list(store).collect()

# <ipython-input-9-2f208d6687d5>:1: RuntimeWarning: coroutine 'AzureAsyncCredentialProvider.__call__' was never awaited
#   obstore.list(store).collect()
# RuntimeWarning: Enable tracemalloc to get the object allocation traceback
# ---------------------------------------------------------------------------
# UnauthenticatedError                      Traceback (most recent call last)
# Cell In[9], line 1
# ----> 1 obstore.list(store).collect()
# 
# UnauthenticatedError: The operation lacked valid authentication credentials for path External Azure credential provider: RuntimeError: no running event loop
# 
# Debug source:
# Unauthenticated {
#     path: "External Azure credential provider",
#     source: PyErr {
#         type: <class 'RuntimeError'>,
#         value: RuntimeError('no running event loop'),
#         traceback: None,
#     },
# }
```

I think that `RuntimeWarning: coroutine 'AzureAsyncCredentialProvider.__call__' was never awaited` is probably the issue.
I think that `UnauthenticatedError: The operation lacked valid authentication credentials for path External Azure credential provider` is a consequence of this, not a cause, because `await credential_provider()` happily returns a token.

kylebarron:
> The async credential provider fails with this error message:

When using async credential providers, you need to use async methods.

Can you try with 
```py
asyncio.run(obstore.list(store).collect_async())
```

docs/authentication.md
```diff
@@ -77,13 +77,13 @@ Refer to [`obstore.auth.boto3`](api/auth/boto3.md).
...
 
 ```py
-from obstore.auth.google import GoogleAuthCredentialProvider
+from obstore.auth.google import GoogleCredentialProvider
 from obstore.store import GCSStore
 
-credential_provider = GoogleAuthCredentialProvider(credentials=...)
+credential_provider = GoogleCredentialProvider(credentials=...)
 store = GCSStore("bucket_name", credential_provider=credential_provider)
 ```
```
kylebarron, line 89: A few lines below, can you uncomment the azure section and an example of using the new `AzureCredentialProvider`/`AzureAsyncCredentialProvider`?

obstore/python/obstore/auth/azure.py
```diff
@@ -0,0 +1,176 @@
...
+
+    ```py
+    from obstore.auth.azure import AzureCredentialProvider
+    from obstore.store import AzureStore
+
+    credential_provider = AzureCredentialProvider(credential=...)
+    store = AzureStore("container", credential_provider=credential_provider)
+    ```
+
+    [`azure.identity`]: https://learn.microsoft.com/en-us/python/api/overview/azure/identity-readme
```
kylebarron, line 79: The [overview page](https://learn.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python) says that the API docs are at `https://aka.ms/azsdk/python/identity/docs`, which redirects to `https://azuresdkdocs.z19.web.core.windows.net/python/azure-identity/latest/azure.identity.html`. Since that's an actual sphinx website, the intersphinx will work. 

Assuming this redirect is stable (`azuresdkdocs.z19.web.core.windows.net` doesn't give me full confidence of a production URL), we can add

```
https://azuresdkdocs.z19.web.core.windows.net/python/azure-identity/latest/objects.inv
```

to the intersphinx config here
https://github.com/developmentseed/obstore/blob/a4f915afe2709aafdddef0485e9348c4c6727d9c/mkdocs.yml#L156-L166

And then remove this literal link.

This should also mean that the type hints interlink. E.g. here: the `Credentials` and `Request` in the signature automatically link to google API docs pages
image

kylebarron:
I'd like to release obstore 0.5 on Monday. If you can just resolve the last two comments, it would be awesome to get this in.

kylebarron:
> Is first thing on Monday morning (AEST) OK?

yeah! That'll be 15 hours before I look at it 😄

docs/authentication.md
daviewales: Done.

obstore/python/obstore/auth/azure.py
daviewales: I figured it out. Just needed to add the API docs: https://github.com/developmentseed/obstore/pull/343/commits/01d644dd7c0dc41c09b6d9ed70a919c1c8e6630b

(It still fails with the `Error reading page` error locally, but at least it's working in CI.)

docs/authentication.md
```diff
@@ -77,19 +77,51 @@ Refer to [`obstore.auth.boto3`](api/auth/boto3.md).
...
 
 Refer to [`obstore.auth.google`](api/auth/google.md).
 
-<!-- #### Using `azure.identity` -->
+#### `azure.identity`
+
+You can use the [`AzureCredentialProvider`][obstore.auth.azure.AzureCredentialProvider] to use [`azure.identity`][] to handle credentials.
+
+``` python
+import obstore
```
kylebarron, line 97:
```suggestion
import obstore as obs
```

docs/authentication.md
```diff
@@ -77,19 +77,51 @@ Refer to [`obstore.auth.boto3`](api/auth/boto3.md).
...
+from obstore.auth.azure import AzureCredentialProvider
+from obstore.store import AzureStore
+
+credential_provider = AzureAsyncCredentialProvider(credential=...)
+store = AzureStore("container", credential_provider=credential_provider)
+print(obstore.list(store).collect())
```
kylebarron, line 103:
```suggestion
print(obs.list(store).collect())
```
I've been suggesting a "canonical" shorter import

docs/authentication.md
```diff
@@ -77,19 +77,51 @@ Refer to [`obstore.auth.boto3`](api/auth/boto3.md).
...
+```
+
+Alternatively, you can use [`AzureAsyncCredentialProvider`][obstore.auth.azure.AzureAsyncCredentialProvider] with the async API:
+
+``` python
+import asyncio
+import obstore
```
kylebarron, line 110:
```suggestion
import obstore as obs
```

docs/authentication.md
```diff
@@ -77,19 +77,51 @@ Refer to [`obstore.auth.boto3`](api/auth/boto3.md).
...
+from obstore.auth.azure import AzureCredentialProvider
+from obstore.store import AzureStore
+
+credential_provider = AzureAsyncCredentialProvider(credential=...)
+store = AzureStore("container", credential_provider=credential_provider)
+
+async def fetch_blobs():
+    blobs = await obstore.list(store).collect_async()
```
kylebarron, line 118:
```suggestion
    blobs = await obs.list(store).collect_async()
```

kylebarron (approved):
This looks great! Thank you!

kylebarron:
I moved the types inline so that they'd show up directly as hyperlinks in the docs: 
image

Issue #342, opened 2025-03-12
Move core exceptions to obspec?


kylebarron:

kylebarron:
I hit a bunch of problems with this in #346 that made me revert obspec entirely, for now, in #367.

Issue #338, opened 2025-03-11
Add obspec as runtime dependency?


kylebarron:
It's cool that it can be a dev-only dependency, and it's not actually _used_ at runtime, but if it's not installed then the user won't get type hints in their editor.

This will be really nice if we get default extras: https://peps.python.org/pep-0771/

kylebarron:
I've added tests that validate that obstore classes are assignable to obspec protocols, so we don't need to make obspec a dependency here.

PR #346, opened 2025-03-13
Use obspec exceptions at runtime

kylebarron:
Closes https://github.com/developmentseed/obstore/issues/342, closes https://github.com/developmentseed/obstore/issues/338. Also closes https://github.com/developmentseed/obstore/pull/199 because the obspec exceptions use multiple inheritance.

Depends on https://github.com/developmentseed/obspec/pull/5

As described below, there's a core issue with this because we'd need to use the exception classes from obspec in pyo3-object_store for the store builders, but we don't want pyo3-object_store to know anything about obspec.


kylebarron:
I got myself into a total mess :melting_face: . I've been trying to do a little refactoring for obstore to make the interface generic, as obspec.

I was looking at the obstore Zarr-Python PR, and I realized we would need to have common exceptions classes, because you might need to have some way to allow a file not found, or something like that. And you'd need to figure out a way to have a common base exception that you can allow.

But then I hit the issue that the base exceptions from obspec would need to be known by the core pyo3-object_store crate, which contains the store builders, because the exceptions defined there would need to subclass from the obspec ones. But I don't really want that because then other rust-python libraries using pyo3-object_store would have to depend on obspec.

So then I thought, I'll manage the exception conversion in the obstore Python layer binding, but then that's super confusing to do.

At the core this is a tug between structural and nominal subtyping. Using protocols instead of subclasses makes this kind of subtyping easier, especially from the Rust side, because the Rust code only has to implement the same shaped interface. The Rust code doesn't need to depend on a specific Python library and subclass from it.

But as far as I know there's no way to do exceptions subtyping with structural subtyping; you can only subclass it.

kylebarron:
Though one thought I just had is that it's ok if only the  "runtime" exceptions subclass from obspec, and not also the "builder" exceptions.

That's maybe a little confusing though.

kylebarron:
It's not simple to handle structural subtyping of exceptions so I'm going to punt on this for the future.

PR #372, opened 2025-03-17
Test Python 3.9 and 3.10 on CI; Apply fixes

kylebarron:
Closes https://github.com/developmentseed/obstore/issues/373

### DIFF
--- a/obstore/python/obstore/_attributes.pyi
+++ b/obstore/python/obstore/_attributes.pyi
@@ -1,4 +1,10 @@
-from typing import Literal, TypeAlias
+import sys
+from typing import Literal
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
 
 Attribute: TypeAlias = (
     Literal[
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -44,7 +44,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     steps:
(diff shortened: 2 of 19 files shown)


obstore/python/obstore/_attributes.pyi
```diff
@@ -1,4 +1,10 @@
-from typing import Literal, TypeAlias
+import sys
+from typing import Literal
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
```
vincentsarago, line 7: You could safely do directly `from typing_extensions import TypeAlias` to avoid checking the python version twice (because typing_extensions` does it as well https://github.com/python/typing_extensions/blob/7def253cd65b3a916fcb76a6eb51428b60e1bcc8/src/typing_extensions.py#L1500-L1517

kylebarron: But people on latest Python versions might not have typing_extensions installed. (we should probably mark it as a runtime dependency ourselves for earlier Python versions I guess)

vincentsarago: > (we should probably mark it as a runtime dependency ourselves for earlier Python versions I guess)

you should 😅 

pydantic adds it for all version https://github.com/pydantic/pydantic/blob/main/pyproject.toml#L46

vincentsarago: FYI it didn't fail on CI because we have a lot of tests dependencies that depends on typing-extensions https://github.com/developmentseed/obstore/blob/main/uv.lock#L281

IMO it doesn't hurt that much to add typing-extensions as a dependency (except that you won't be able to say that obstore is a zero dependency module 😬)

kylebarron: #375

PR #432, opened 2025-04-25
fix: docs for json.loads(bytes)

gadomski:
The docs indicate that you can `json.loads(bytes)` but you can't. I've ~added a failing test to demonstrate, and~ corrected the docs. ~Not sure if we _should_ be able to `json.loads`?~

### DIFF
--- a/obstore/python/obstore/store.py
+++ b/obstore/python/obstore/store.py
@@ -551,7 +551,8 @@ class HTTPStore(_ObjectStoreMixin, _store.HTTPStore):
     store = HTTPStore.from_url("https://api.github.com")
     resp = obs.get(store, "repos/developmentseed/obstore")
 
-    # If you have orjson installed, you can load bytes without copying them via `memoryview`:
+    # If you have orjson installed, you can load bytes without copying them via
+    # `memoryview`:
     import orjson
     data = orjson.loads(memoryview(resp.bytes()))
 


kylebarron:
Interesting. I swear I tried this before adding it to the docs but I can confirm that even testing on 3.13.2 this fails. Ref https://github.com/python/cpython/issues/74379

Should we change to suggesting orjson instead? Orjson works with memoryviews (without copying to `bytes`) but it doesn't work with arbitrary buffer protocol objects yet: https://github.com/ijl/orjson/issues/562

gadomski:
I think a **orjson** suggestion while still describing how to fall back to **json** would work. E.g.

> If you have **orjson**, you can work with bytes directly:
>
> ```pyhthon
> data = orjson.load(b)
> ```
>
> If you don't, you'll need to create `bytes` first, which is a bit less efficient:
>
> ```python
> data = json.load(bytes(b))
> ```

kylebarron:
Yes, that looks great, except today with orjson you still have to manually wrap it in `orjson.loads(memoryview(obstore_bytes))`, even though that `memoryview` is zero-copy.

obstore/python/obstore/store.py
```diff
@@ -550,7 +550,13 @@ class HTTPStore(_ObjectStoreMixin, _store.HTTPStore):
 
     store = HTTPStore.from_url("https://api.github.com")
     resp = obs.get(store, "repos/developmentseed/obstore")
-    data = json.loads(resp.bytes())
+
+    # If you have orjson, you can load bytes without copying them via `memoryview`:
```
kylebarron, line 554:
```suggestion
    # If you have orjson installed, you can load bytes without copying them via `memoryview`:
```

kylebarron:
Thanks!

kylebarron:
haha you managed to push another commit before it auto-merged 😛

Issue #403, opened 2025-04-03
Make example with Azure planetary computer Parquet data with fsspec store


kylebarron:
```
collection = catalog.get_collection("landsat-c2-l2")
asset = collection.assets["geoparquet-items"]

credential_provider = PlanetaryComputerCredentialProvider(
    asset.href,
    account_name=asset.extra_fields["table:storage_options"]["account_name"],
)
```

Make example of using this data with a pyarrow Dataset

kylebarron:
See example in https://github.com/developmentseed/obstore/issues/413#issuecomment-2784251040, even if that's not intended behavior

PR #433, opened 2025-04-28
Add example for Planetary Computer GeoParquet with PyArrow

demoncoder-crypto:
Addresses #403

### DIFF
--- a/examples/planetary_computer_geoparquet.py
+++ b/examples/planetary_computer_geoparquet.py
@@ -1,0 +1,129 @@
+import pystac_client
+import pyarrow.dataset as ds
+import obstore
+import obstore.auth.planetary_computer
+import os
+
+# Optional: Set environment variable if needed for Azure access
+# os.environ["AZURE_STORAGE_ACCOUNT_NAME"] = "..." # Usually picked up from asset
+
+# URL for the Microsoft Planetary Computer STAC API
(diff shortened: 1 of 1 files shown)


demoncoder-crypto:
I have done the basic implementation if its correct I will fix everything iteratively. Thanks

examples/planetary_computer_geoparquet.py
```diff
@@ -0,0 +1,129 @@
...
+    )
+    print(f"AzureStore initialized for account: {account_name}, URL base: {store_url}")
+else:
+    print("Error: Cannot initialize AzureStore without account_name.")
+    exit()
+
+
+# Create an fsspec filesystem interface from the obstore instance
+print("Creating fsspec filesystem from obstore...")
+fs = obstore.fsspec.ObstoreFileSystem(azure_store)
```
kylebarron, line 92: This doesn't exist.

kylebarron (changes requested):
This is an extremely verbose (AI-generated?) example where we really only need 10 lines of code, such as what we already have in the Zarr Planetary Computer example: https://github.com/developmentseed/obstore/blob/50782ed782a15185a936d435d13ca0a7969154ae/examples/zarr/main.py#L11-L24

We already have an example for creating the Planetary Computer credential provider, so it would be much better to have a focused example on how to use pyarrow than a verbose example with all these print statements.

PR #441, opened 2025-05-09
Remove our own tokio runtime and use the one provided by pyo3-async-runtimes

kylebarron:
…untimes

## Available PR templates


kylebarron:
> I believe the forked processes checking remains, pyo3-async also uses Tokio right?

pyo3-async-runtimes does use tokio, but I don't think it has as nice of a check as was here. I figure tokio will just silently stop working if the python user tries to multiprocess?

ion-elgreco:
> > I believe the forked processes checking remains, pyo3-async also uses Tokio right?
> 
> pyo3-async-runtimes does use tokio, but I don't think it has as nice of a check as was here. I figure tokio will just silently stop working if the python user tries to multiprocess?

It might deadlock or panic actually

kylebarron:
Indeed this might fail when using fork-based multiprocessing, but we can wait to figure out how to document that when the first user complains.

kylebarron:
I think that yes, obstore is not fork safe. Do you have a suggestion of where to document that?

Issue #392, opened 2025-03-25
Earthdata credential provider improvements


kylebarron:
- [ ] Check for environment variables
- [ ] Check in .netrc
- [ ] Don't hard-code token provider URL
- [ ] Support bearer tokens
- [ ] Don't use requests.session object.

Ref https://github.com/nsidc/earthaccess/discussions/969

betolink:
One thing I'm not sure about the list above is the `Don't use requests.session object.` In earthaccess we use a sub-classed `requests.session` object to handle the OAuth2 redirect. I'll take a look at how to implement the Earthdata credential provider!

kylebarron:
In https://github.com/nsidc/earthaccess/discussions/969#discussioncomment-12617721 @chuckwondo said

> I also suggest _not_ using a session object for 2 major reasons:
> 
> 1. You're using the session in an insecure manner because you're leaking credentials (i.e., supplying the creds on requests that are _not only_ to EarthData Login)
> 2. The sessions are resource leaks because they are not reliably/consistently closed (because it's not clear when or if the store object using them will ever close the provider, which would close the session).
> 
> In other words, it appears that you're using a session to simplify the code, but that simplification not only leaks resources, but also creates a security hole.

I think that's a fair critique and am happy to not use a `Session`

kylebarron:
See draft PR in #472

PR #454, opened 2025-05-17
Bump object-store to 0.12.1

chuckwondo:
## What I am changing

- bumping `object-store` dependency version to `0.12.1`, which was released May 9

This addresses the following error, which appears to be due to 0.12.1rc no longer being available?

```plain
$ uv run maturin develop -m obstore/Cargo.toml
error: failed to load source for dependency `object_store`

Caused by:
  Unable to update https://github.com/apache/arrow-rs-object-store?rev=ed9329f28eba5b1e0f6bbd10c79738ec42a9f568#ed9329f2

Caused by:
  failed to map 'no'; class=Config (7)
💥 maturin failed
  Caused by: Cargo metadata failed. Does your crate compile with `cargo build`?
  Caused by: `cargo metadata` exited with an error: 
```

## How I did it

- updated `Cargo.toml`

## How you can test it

- Run `uv run maturin develop -m obstore/Cargo.toml` to verify the above error does _not_ occur
- Run `cargo test --all` to verify that all tests pass

@kylebarron, in preparation to work on #392, I ran into this problem, but I don't know if the change I made here is the proper way to fix it (i.e., the fix works, but I don't know if this is the "correct" way to fix it).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1394,7 +1394,8 @@ dependencies = [
 [[package]]
 name = "object_store"
 version = "0.12.1"
-source = "git+https://github.com/apache/arrow-rs-object-store?rev=ed9329f28eba5b1e0f6bbd10c79738ec42a9f568#ed9329f28eba5b1e0f6bbd10c79738ec42a9f568"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


kylebarron (approved):
Yeah this looks great! (I'm surprised the git sha is no longer there)

PR #472, opened 2025-05-27
feat: Improve NASA Earthdata credential providers

### LINKED ISSUE #392: Earthdata credential provider improvements

chuckwondo:
## What I am changing

- Fixing and completing NASA Earthdata credential providers

## How I did it

- Fixed/enhanced NasaEarthdataCredentialProvider
- Fixed/enhanced NasaEarthdataAsyncCredentialProvider

## How you can test it

- TBD

## Related Issues

- Fixes #392


chuckwondo:
Oops! I see a couple of failed checks. I'll look at them tomorrow.

obstore/python/obstore/auth/_http.py
```diff
@@ -0,0 +1,67 @@
...
+from warnings import warn
+
+if TYPE_CHECKING:
+    import aiohttp
+    import aiohttp_retry
+    import requests
+
+
+def default_requests_session() -> requests.Session:
+    # Upstream docstring in case we want to expose these values publicly
```
kylebarron, line 13: I think this specifically means upstream in planetary-computer code, right?

obstore/python/obstore/auth/earthdata.py
```diff
@@ -6,157 +6,419 @@
...
+    import aiohttp
+    import aiohttp_retry
+    import requests
+
     from obstore.store import S3Config, S3Credential
 
-CREDENTIALS_API = "https://archive.podaac.earthdata.nasa.gov/s3credentials"
+
+EARTHDATA_HOST_OPS = "urs.earthdata.nasa.gov"
+EARTHDATA_HOST_UAT = "uat.urs.earthdata.nasa.gov"
```
kylebarron, line 28: Can we have docstrings for each of these constants, which will get rendered in the docs?

We'll also need to add these constants to the docs here: https://github.com/developmentseed/obstore/blob/6d878a2e39e5bc26f87ce8505fac821b42f2689e/docs/api/auth/earthdata.md?plain=1#L1-L8

obstore/python/obstore/auth/earthdata.py
```diff
@@ -6,157 +6,419 @@
...
+            msg = "credential provider was closed"
+            raise ValueError(msg)
+
+        metadata = await self._get_credentials_async(self._session)
+
+        return _metadata_to_s3_credential(metadata)
 
     async def close(self) -> None:
-        """Close the underlying session.
+        """Close the underlying session, if it was not supplied by the user.
```
kylebarron, line 312: Are there any external references we can consider here? I'm inclined to make `close` _always_ close the underlying session, and the user can decide not to call `close` if they want to keep using the underlying session. Seems like that would be more predictable.

obstore/python/obstore/auth/earthdata.py
```diff
@@ -6,157 +6,419 @@
...
-        # Pass region default
+        """
         self.config = {"region": "us-west-2"}
-        self.session = ClientSession(auth=BasicAuth(username, password))
+        self._session: aiohttp.ClientSession | aiohttp_retry.RetryClient | None = (
+            session or default_aiohttp_session()
+        )
+        # Avoid closing a user-supplied session (the user is responsible for that)
+        self._close = None if session else self._session.close
+        self._get_credentials_async = _make_get_credentials_async(
```
kylebarron, line 295: Instead of a factory function that returns another function; can we just move the body of `def _make_get_credentials_async` into a class instance method? Seems like that would be a lot simpler and we wouldn't need to deal with closures.

obstore/python/obstore/auth/earthdata.py
```diff
@@ -6,157 +6,419 @@
...
+        r.raise_for_status()
+
+        try:
+            return r.json()
+        except json.JSONDecodeError:
+            # Content is not JSON; basic auth creds are invalid or not found via netrc
+            r.status_code = 401
+            r.reason = "Unauthorized"
+            r.raise_for_status()
+            return {}  # Never reached, but necessary for type checking
```
kylebarron, line 222: Nit: I usually prefer an `assert False` to make clear the code is unreachable. I think that also makes type checking pass?

obstore/python/obstore/auth/earthdata.py
```diff
@@ -6,157 +6,419 @@
...
+        underlying [requests.Session][].  If you supplied your own session object,
+        it will not be closed.
         """
-        self.session.close()
+        if self._close:
+            self._session = None
+            self._close()
+
+
+def _make_get_credentials(
```
kylebarron, line 163: Ditto about the suggestion to make this a class instance method instead of a factory function.

obstore/python/obstore/auth/earthdata.py
chuckwondo: Sure thing. I'm not familiar with the required syntax in the md file. Should I simply add 2 more entries similar to those 2 existing entries in the md file?

obstore/python/obstore/auth/earthdata.py
chuckwondo: Ah, sure, let me see.

maxrjones:
> > preferentially chosen over the sync version

> Preferentially chosen... by the IDE autocompletion?

My assumption is that most autocompletion implementations use alphabetical order rather than random selection but I could be wrong.

obstore/python/obstore/auth/earthdata.py
```diff
@@ -6,157 +6,434 @@
...
-CREDENTIALS_API = "https://archive.podaac.earthdata.nasa.gov/s3credentials"
+
+EARTHDATA_HOST_OPS = "urs.earthdata.nasa.gov"
+"""Hostname of the NASA Earthdata Login Operational (OPS) system.
+
+This is the default host used during credential authorization.
+"""
+
+EARTHDATA_HOST_UAT = "uat.urs.earthdata.nasa.gov"
+"""Hostname of the NASA Earthdata Login User-Acceptance Testing (UAT) system."""
```
kylebarron, line 39: Can you say when a user might want to use this instead?

Is this only for testing? If it's not intended for end-user use, we should remove it from here.

obstore/python/obstore/auth/earthdata.py
```diff
@@ -6,157 +6,434 @@
...
+        https://archive.podaac.earthdata.nasa.gov/s3credentialsREADME
+
+    Examples:
+        ```py
+        import obstore
+        from obstore.auth.earthdata import NasaEarthdataCredentialProvider
+
+        # Obtain an S3 credentials URL and an S3 data/download URL, typically
+        # via metadata returned from a NASA CMR collection or granule query.
+        credentials_url = "https://data.ornldaac.earthdata.nasa.gov/s3credentials"
```
kylebarron, line 75: It might be good to also have an initial example that doesn't pass a custom credentials url? It's not clear to me when a user would need to override this url, and so I could see a user getting confused by an example that passes this, even though the argument has a default.

kylebarron:
I made some small edits; take a look.

> Let me know if you'd like me to add them to this PR or submit a separate PR.

Maybe put those in a new PR? Might be easier to follow just the test and CI changes then.

obstore/python/obstore/auth/earthdata.py
chuckwondo: This is not often used, but it is something that may still want to expose to users. Users of the `earthaccess` library have expressed the desire to hit "test" data rather than "production" data, and both hosts are coded into that library.

However, I'm not opposed to removing this. Users who want to hit the UAT system will know the host name and can simply provide it explicitly themselves, so I'd be okay with having only the "ops" host, in which case, I might be inclined to simply name it `DEFAULT_EARTHDATA_HOST`.

What do you prefer?

obstore/python/obstore/auth/earthdata.py
kylebarron: Can we set a default? Is there a DAAC that's most commonly used? Why does `earthaccess.login` not require the user to pass a URL? https://earthaccess.readthedocs.io/en/latest/tutorials/file-access/

kylebarron:
cc @betolink, @mfisher87 in case you have any comments.

obstore/python/obstore/auth/earthdata.py
chuckwondo: The logic in `earthaccess` that is handing the "endpoint" (credentials URL) is equivalent to the logic that's in this obstore credential provider, but the code in the PR adds further functionality by supporting the use of a token as well as recognizing env vars.

kylebarron: I see previously the code I wrote hardcoded 
```py
CREDENTIALS_API = "https://archive.podaac.earthdata.nasa.gov/s3credentials"
```

Is there a known set of DAACs? Could we have these on an enum so the user doesn't need to know a full URL?

```py
class EarthdataProviders(Enum):
    PODAAC = "https://archive.podaac.earthdata.nasa.gov/s3credentials"
...
```
?

kylebarron (approved):
Thank you!

obstore/python/obstore/auth/earthdata.py
chuckwondo: > I want this to be useful _not only_ to users who are already using the `earthaccess` library (in that case, we could have this entire file live in the earthaccess project).

Yes, I absolutely agree with you here!

> 
> > The user doesn't need to know these URLs. The URLs come as part of the metadata returned by a CMR search.
> 
> I suppose that means a user will need separate `S3Store` objects for different datasets? And they won't be able to construct an `S3Store` until doing a CMR search because they won't be able to construct the credential provider until seeing the CMR result...

We need an S3Store per provider/DAAC (which typically equates to a bucket -- I'm not aware of a DAAC using multiple buckets for data).

For the most part, each DAAC will host multiple datasets (collections, not to be confused with the STAC term), and will use the same credentials URL across all of their hosted datasets. However, I don't think it's safe to assume this. A DAAC could very well partition their datasets such that a user who may have access to some datasets, may not necessarily have access to other datasets, thus their may be the possibility of multiple credentials URLs per DAAC. Offhand, I don't now if any DAAC currently does this, so this could be an extreme edge case.

Also, most commonly, a user makes a CMR query for data from a single collection, so all results would use the same credentials URL (most likely). However, it would not be strange to issue a CMR query that spans multiple datasets, or even multiple DAACs, thus requiring multiple credentials URLs (multiple credential providers). I can even imagine doing this across GEDI data, where I might want to perform a join of 2 GEDI datasets, but 1 is hosted by LP DAAC and the other is hosted by ORNL DAAC.

Generally speaking, yes, they won't be able construct a store until they do a CMR search and obtain the metadata, but that is the general workflow with NASA Earthdata datasets, regardless of the library being used (unless they're using a library specialized for a particular DAAC's holdings, such as ASF, which hides all of these details).

You're likely going to be hard-pressed to find anybody holding onto a list of bare S3 URLs for data files, which would then also require them to maintain a list of associated credentials URLs (or at least 1 credentials URL, if all S3 URLs are for the same DAAC/bucket).

The common use case for accessing NASA Earthdata data is to perform a CMR query and extract the S3 data URL along with the related credentials URL from the response, for each item in the result. I've never seen an example usage that doesn't do this. Of course, that doesn't mean they don't exist, but they would be rather odd (unless somebody has simply hardcoded an S3 URL and its related credentials URL simply to use as an example that wouldn't require a CMR query, simply for expediency for the example, exactly what I've done in the examples in the docstrings).

However, I have some thoughts about things to explore for potentially making things easier for the user, whether or not they're using `earthaccess`, so I'm going to open a new discussion. There's definitely much room for improvement.

PR #478, opened 2025-06-02
fix: Fix parsing Azure bucket URL

kylebarron:
## Available PR templates


kylebarron:
We may choose to just use the upstream fix https://github.com/apache/arrow-rs-object-store/issues/398 instead

PR #487, opened 2025-06-23
feat: Deprecate custom NotFoundError in favor of built-in FileNotFoundError

kylebarron:
## Available PR templates

### DIFF
--- a/obstore/pyproject.toml
+++ b/obstore/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "maturin"
 [project]
 name = "obstore"
 requires-python = ">=3.9"
-dependencies = ["typing-extensions; python_version < '3.12'"]
+dependencies = ["typing-extensions; python_version < '3.13'"]
 classifiers = [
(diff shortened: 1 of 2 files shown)


lukasbindreiter:
The only small downside with fully using `builtins.FileNotFoundError` I think is it's name - since no files are necessarily involved.

```python
from obstore.store import MemoryStore

MemoryStore().get("non-existent")  # FileNotFoundError
# no files involved anywhere
```

I guess that's why the exception in obstore was just named `NotFoundError`?
For the above example `KeyError` would probably be the more appropriate, but then for other stores not really.

kylebarron:
All stores will raise the same exceptions, that's for sure. Having different stores raise different exceptions would be so confusing IMO

lukasbindreiter:
> Having different stores raise different exceptions would be so confusing IMO

yeah for sure, totally agree with that. Just wanted to mention that `FileNotFoundError` itself may also be confusing in certain scenarios - but for now it is the best approach for sure.

PR #501, opened 2025-07-17
chore: Bump object_store to 0.12.3

kylebarron:
Upstream changelog is here: https://github.com/apache/arrow-rs-object-store/blob/main/CHANGELOG.md#v0123-2025-07-11

Should include portions of that in the obstore changelog?


kylebarron:
Merging to unblock ci for #504. Will update docs separately

PR #505, opened 2025-07-18
Fix pyright config

pjonsson:
Use correct syntax for ignoring
directories, set a stubPath,
and configure the environment
so pyright can find the modules.

This discovered a type error in
conftest.py, so fix that as well.

Here's what a locally installed pyright run looks like on top of this PR:
```
$ uv run pyright
.../obstore/obstore/python/obstore/__init__.py
  .../obstore/obstore/python/obstore/__init__.py:4:6 - warning: Import "._obstore" could not be resolved from source (reportMissingModuleSource)
.../obstore/obstore/python/obstore/store.py
  .../obstore/obstore/python/obstore/store.py:8:6 - warning: Import "obstore._obstore" could not be resolved from source (reportMissingModuleSource)
  .../obstore/obstore/python/obstore/store.py:9:6 - warning: Import "obstore._obstore" could not be resolved from source (reportMissingModuleSource)
  .../obstore/obstore/python/obstore/store.py:35:10 - warning: Import "obstore._obstore" could not be resolved from source (reportMissingModuleSource)
0 errors, 4 warnings, 0 informations
```


pjonsson:
> should be `pull_request_target`, not `pull_request`. Would you be able to change that?

I'm pretty sure that change needs to be merged to main before this PR can go green, so I put that in #508.

I don't understand these parts of the permission model in Github, so from my perspective it's quite possible the consequence of that PR being merged will be that a crocodile eats your bedpost.

pjonsson:
> > Edit: I believe it's easiest to just ignore `reportMissingModuleSource` for the `_obstore` imports, there isn't supposed to be a Python source file for that module.
> 
> Is there a way to do that at the `pyproject.toml` level like mypy does?

Sure, it's just to chuck in `reportMissingModuleSource = false` in the pyright section of `pyproject.toml`. Do you want me to put it there and remove the ignores that I added in this PR? (That will disable the check for all sources though, but I assume you already realize the tradeoffs with that.)

pjonsson:
> mypy allows missing type hints _only_ for the `fsspec` module. Does pyright not have a per-module config like that?

We're deep into the weeds of Pyright now and I'm just a regular user, so I don't know. If you look towards the end of issue 4059 in the Pyright repository, I get the impression they have sliced the problem in a different direction.

I tested defining an environment that set `_obstore.pyi` as its root and disabled the warning, but that still warned on the imports in `__init__.py`. I also tried adding a comment at the top of `_obstore.pyi` that disabled the warning, but it seems the warning needs to be disabled for the importing module.

If one needs to disable a warning for the entire importing module, I would recommend having individual ignores on the lines that need them instead so the rest of the module can benefit from those warnings. The number of imports needing the ignores is small, and the problem appears when type checking obstore itself, I don't think users of obstore will get the warnings. Additionally, the Pyright ignore annotations are not recognized by MyPy and other type checkers, so it will just be a regular comment to those.

kylebarron (approved):

kylebarron:
Yes, that would be great; thank you!

PR #524, opened 2025-08-04
fix: Support percent-encoded paths

kylebarron:
## Available PR templates


ion-elgreco:
@kylebarron yeah this is the correct way, we shouldn't encode provided paths, but assume the provided path is correct, this should be handle the end-user of the library.

Also ran some tests and this indeed fixes some issue we ran into. So we should use Path::parse indeed

kylebarron:
Thanks for the feedback!

kylebarron:
Released in https://github.com/developmentseed/obstore/releases/tag/py-v0.8.0

ion-elgreco:
Thanks @kylebarron

PR #544, opened 2025-08-27
Add instrumented object store to pyo3-object-store

### LINKED ISSUE #93: Add logging
See https://pyo3.rs/v0.22.5/ecosystem/logging, https://github.com/vorner/pyo3-log.

It looks like upstream is already logging: https://github.com/apache/arrow-rs/blob/2b373f850c5ac3d8ba6a1a0dfb98f5711f5318c8/object_store/src/aws/builder.rs#L939

kylebarron:
### Change list

- Adds support for logging 
- Vendors `instrumented_object_store` from https://github.com/datafusion-contrib/datafusion-tracing/blob/main/instrumented-object-store/src/instrumented_object_store.rs. If we go ahead with this approach, I'll make a PR there upstream with a couple changes (allowing `T: ObjectStore` instead of `Arc<dyn ObjectStore>` and adding an `inner(&self) -> &T` method).
- Adds Python `init_log` function for creating a global `tracing` subscriber.
- This `init_log` function can only be called once because it initiates into a 
    - Perhaps we could instead have a `Logger` class that holds the `guard` and avoids dropping it? We don't necessarily need to instantiate the tracing subscriber into global scope.
    - This would also allow multiple logging outputs to different locations based on trace/debug/etc level.
- For now, it's **only** possible to log to files. I think this is simpler (and more performant) than trying to log back into Python because otherwise we get GIL contention across the threads.
- The tracing store wrapper is _always_ used, whether or not logging has been requested. This simplifies code a lot (otherwise we'd need to store `enum`s of whether the user has requested logging or not). I _think_ that the performance impact will be negligible when logging has not been opted-into via `init_log`.

Closes #93


kylebarron:
Latest commits let you mix and match stdout/stderr/file, so you can choose different configurations for each output destination

This config:
```py
stderr_config = {
    "format": "json",
    "show_ansi": False,
    "show_target": False,
    "show_level": False,
}
stdout_config = {
    "format": "pretty",
    "show_ansi": True,
    "show_target": True,
    "show_level": True,
}
init_log(stderr=stderr_config, stdout=stdout_config, level="trace")
```
produces:
image

kylebarron:
And I added support for destination-specific levels, so here stdout gets only debug events while stderr gets trace and debug events:

```py
stderr_config = {
    "format": "json",
    "show_ansi": False,
    "show_target": False,
    "show_level": False,
    "level": "trace",
}
stdout_config = {
    "format": "pretty",
    "show_ansi": True,
    "show_target": True,
    "show_level": True,
    "level": "debug",
}
```

image

kylebarron:
@geospatial-jeff Should I support file-based logging still? Or do you think I should only give an option of stderr and stdout?

geospatial-jeff:
No just do stderr and stdout, let end user decide where to send the logs.

aabed-aa:
I think integrating with https://www.structlog.org/en/stable/ would be great

PR #545, opened 2025-08-29
feat: Paginated list support to allow substring list prefix

### LINKED ISSUE #494: Wildcard prefix in the list command
Hello,

I have tons of log files with data at the beginning for example `202506272215_blabla` on S3 I can use prefix as substring, basically I can get all files for this day by `my_folder/20250627*` but it's not working in obstore.

How I can get all files by prefix and not by folder path?

kylebarron:
Currently, obstore _only_ supports listing by path segments. So if you pass a prefix into `list_with_delimiter` or `list`, that will be assumed to be a full path segment. This means that it's currently impossible to do efficiently perform the desired query from #494:

> I have tons of log files with data at the beginning for example `202506272215_blabla` on S3 I can use prefix as substring, basically I can get all files for this day by `my_folder/20250627*` but it's not working in obstore.

`object_store` supports substring-based prefix listing in its [`PaginatedListStore`](https://docs.rs/object_store/latest/object_store/list/trait.PaginatedListStore.html) API.

### Change list

This PR changes to use the `PaginatedListStore` for all `list` operations, which then support substring-based prefix filters.

However, this `PaginatedListStore` is only implemented for S3, Azure, and GCS. It's not implemented for HTTPStore or LocalStore, because those don't have a concept of pagination. See https://github.com/apache/arrow-rs-object-store/issues/388.

We use an emulated `list`, which materializes all paths in the directory in memory first, for LocalStore and HTTPStore. It's not a problem for HTTPStore because the underlying client for WebDAV materializes the full list anyways. And the LocalStore is local and lists should be fast.

Closes https://github.com/developmentseed/obstore/issues/494


kylebarron:
The latest two commits added an implementation of substring-match `list`, partially written by Claude.

- Fix clippy lints
- Add test using minio? I need to have a python test that runs on the paginated implementation too
- Propagate error through the stream https://github.com/developmentseed/obstore/blob/958fd4d68ffd896d05c53828ba92112426f148f8/obstore/src/list.rs#L562
- Make `create_filtered_stream` more concise. Did I write a helper for that initially?
- Use set equality for paths in Python tests

kylebarron:
### Re: "implement Local/HTTP via a stream transform so we never materialize the entire stream"

I investigated whether the emulated path (`emulate_paginated_list`, used for `MemoryStore` / `LocalStore` / `HttpStore`) should stream results instead of collecting them into memory. **Conclusion: not worth it — keep the materialize-then-filter approach and the unified `MaybePaginatedStore` abstraction.**

Avoiding materialization only matters for a *remote + potentially large* listing, and none of the three emulated backends fit that:

| Store | `list` backing | Materialized regardless? |
|---|---|---|
| Memory | in-memory map | yes — already in RAM |
| Local | filesystem walk | local + cheap (`ObjectMeta` is tiny) |
| HTTP | `PROPFIND` `Depth: infinity` → one `<multistatus>` XML doc | **yes, by the protocol** |

The HTTP case is the crux. `HttpStore::list` is WebDAV: it issues a single `PROPFIND` with `Depth: infinity`, and object_store deserializes the **entire** response into a `Vec` before wrapping it in `stream::iter`. So "never materialize the HTTP stream" is unachievable — the data is fully in memory the moment the response is parsed. On top of that, WebDAV stores are rare in practice; most `HttpStore` usage is a plain `GET` against an HTTP server/CDN, where `PROPFIND` returns `405` and `list` doesn't function at all.

So a streaming rewrite would add a second code path (and likely dissolve `MaybePaginatedStore`) to solve a problem that doesn't exist for any emulated backend. The unified abstraction with materialize-then-filter is simpler and correct.

*— Analysis by Claude (Claude Code), posted at @kylebarron's request*

PR #547, opened 2025-09-02
chore(pyo3-bytes): Support pyo3 0.26

kylebarron:
## Available PR templates


kylebarron:
Yes yes sorry; I got distracted with lunch and a dr appt; it's on crates.io now!

PR #557, opened 2025-09-08
demonstrate LocalStore.head() failing on Windows OS

keen85:
`LocalStore.head()` fails for existing local folders in Windows


kylebarron:
According to upstream (https://github.com/apache/arrow-rs-object-store/issues/483) this is expected not to work.

PR #578, opened 2025-10-07
fix: Remove TypeVar constraints on arro3-core to fix list typing when arro3-core not installed

kylebarron:
## Available PR templates


kylebarron:
@ion-elgreco can you test out this branch to see if it works for you as well?

ion-elgreco:
@kylebarron great, this indeed works! So when arro3-core is not installed and return_arrow=False, you properly see Sequence[objectMeta]. If you set return_arrow=True it gives unknown, but that makes sense.

Issue #589, opened 2025-10-29
S3Store.from_url fails with https URL


kylebarron:
```
        # ("https://nasa-power.s3.amazonaws.com/syn1deg/temporal/power_syn1deg_monthly_temporal_lst.zarr", {}),
```

ref https://github.com/developmentseed/titiler/pull/1253#discussion_r2472827432

cc @vincentsarago

PR #590, opened 2025-10-29
feat: Allow S3 HTTP URLs without region

kylebarron:
Closes https://github.com/developmentseed/obstore/issues/589

This doesn't infer a region from the URL by default. I think `S3Store` defaults to `us-east-1` if no other information has been passed. So for any URLs in other regions, the user will have to specify the correct region.

cc @vincentsarago 

We should make this same PR to upstream as well


kylebarron:
Ref similar upstream PR in https://github.com/apache/arrow-rs-object-store/pull/523 (we vendor this piece of code so that we can apply URL parsing onto the _config_ instead of onto the _builder_)

PR #593, opened 2025-11-07
fix: Prevent early EOF error in reader.read

### LINKED ISSUE #592: Reader raises after EOF
Calling `read` with nonzero size on Python file-likes return an [empty byte](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read) string once EOF is reached.

It seems obstore readers are inconsistent with this behavior and raise an exception instead.

```python

f = open(..., "rb")
f.read()
f.read(1)  # -> b""

f = fsspec.open(..., mode="rb").open()
f.read()
f.read(1)  # -> b""

store = obstore.store.from_url(..., skip_signature=True)
f = obstore.open_reader(store, "")
f.read()
f.read(1)  # OSError: early eof
```

nvictus:
Fixes #592 

Now `reader.read(size)` will return a truncated byte string instead of raising an OSError when `size` would extend past the end of the file, mirroring the behavior of the standard library IO types.

### DIFF
--- a/tests/test_buffered.py
+++ b/tests/test_buffered.py
@@ -1,3 +1,5 @@
+from io import BytesIO
+
 import pytest
 
 import obstore as obs
@@ -90,6 +92,10 @@ def test_read_past_eof_sync():
     buffer = file.read(20)
     assert memoryview(data) == memoryview(buffer)
 
+    buf = BytesIO(data)
+    expected = buf.read(20)
+    assert memoryview(expected) == memoryview(buffer)
+
 
 @pytest.mark.asyncio
 async def test_read_past_eof_async():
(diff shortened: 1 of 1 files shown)


tests/test_buffered.py
```diff
@@ -77,3 +77,28 @@ async def test_writable_file_async():
     resp = await obs.get_async(store, path)
     retour = await resp.bytes_async()
     assert retour == line * 50
+
+
+def test_read_past_eof_sync():
+    store = MemoryStore()
```
kylebarron: Can you also compare this with `BytesIO` so that we know we're matching stdlib behavior? Like
```
buf = BytesIO(data)
expected = buf.read(20)
```

nvictus: done!

PR #613, opened 2026-02-04
Add import statement for obstore in examples

jbusecke:
Just ran into an issue with the example here. I think we might also be able to replace
```
obstore.get(store, filename) 
# with
store.get(filename)
```
?


kylebarron:
Would you be able to instead fix this example by switching to `store.get`?

PR #620, opened 2026-02-19
feat: Update to pyo3 0.28; pin to <3.14 for now.

DisturbedOcean:


kylebarron:
The failing type check might be a new type check that pyright added. You should be able to put a `type: ignore` on that function line to fix it

kylebarron (approved):
Thanks!

PR #658, opened 2026-04-03
ci: enhance dependabot config to track pip and cargo packages

lhoupert:
Follow up from #647. Adding pip and cargo package to dependabot that should help quickly patch vulnerabilities in dependencies package


.github/dependabot.yml
```diff
@@ -13,3 +13,73 @@ updates:
...
+    groups:
+      minor-and-patch:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "uv"
```
kylebarron: What's the difference between package ecosystem `uv` and `pip`?

.github/dependabot.yml
```diff
@@ -13,3 +13,73 @@ updates:
...
+      interval: "weekly"
+      day: "monday"
+    groups:
+      minor-and-patch:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
```
kylebarron: pyo3-bytes and pyo3-object_store are libraries that form part of this monorepo and are not applications. 

There's only one top-level Cargo.lock for the entire monorepo, so I'm not sure whether you need these here?

.github/dependabot.yml
lhoupert: Do you have a preference on which one to use @kylebarron ?

kylebarron:
I removed `uv` and `pip` because the applications in obstore are for examples. They're not deployed and we don't need to keep them up to date

PR #663, opened 2026-04-10
fix(fsspec): _info() should honor self.dircache

fvaleye:
# Summary
This PR eliminates unnecessary network calls in `FsspecStore._info()` by ensuring it consults the directory cache before falling back to a network request, bringing it into alignment with standard fsspec implementations.

Currently, `FsspecStore._info()` issues an unconditional `head_async()` and ignores `self.dircache`, even though [`AbstractFileSystem._ls_from_cache`](https://github.com/fsspec/filesystem_spec/blob/0bb819cf923b8680ac473f6a2faf492ab4d3602f/fsspec/spec.py#L364) is documented as the standard way for callers to surface pre-known file metadata. 

For comparison, both [s3fs](https://github.com/fsspec/s3fs/blob/731e1250bcd4f682e1ccce03b01641910e7646fc/s3fs/core.py#L1698-L1730) and [gcsfs](https://github.com/fsspec/gcsfs/blob/64936ae72aa99024282189333b476f0875bbb0ca/gcsfs/core.py#L1048-L1080) consult `_ls_from_cache()` first and only fall back to a network call on miss. `obstore.fsspec.FsspecStore` is the outlier here: pre-populating `fs.dircache` (or having it populated by a prior `_ls`) is ignored.

# Notes
When the parent directory is cached, and the requested child is not in the listing, `_info()` now raises `FileNotFoundError` immediately instead of falling through to a `_ls()`. This matches `s3fs._info` and the documented contract of `_ls_from_cache`: a cached parent listing is authoritative, so a missing child definitively does not exist.

### DIFF
--- a/obstore/python/obstore/fsspec.py
+++ b/obstore/python/obstore/fsspec.py
@@ -487,6 +487,16 @@ class FsspecStore(fsspec.asyn.AsyncFileSystem):
         await self._local_store._pipe_file(lpath, resp)  # noqa: SLF001
 
     async def _info(self, path: str, **_kwargs: Any) -> dict[str, Any]:
+        # Consult `self.dircache` before issuing a HEAD request. An empty
+        # filter means the cache has evidence `path` is a directory but no
+        # entry under that exact name, so we return a synthetic directory.
+        cached = self._ls_from_cache(path)
+        if cached is not None:
+            match = next((entry for entry in cached if entry["name"] == path), None)
+            if match is not None:
+                return match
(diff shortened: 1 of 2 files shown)


kylebarron (approved):
Thank you!

This is why fsspec is so awful to work with. Each implementation needs to know specifics of other internals of fsspec to be correct. Even if fsspec decides to implement caching, it should do it at a level that doesn't require implementors to know about it!

(FWIW I'm happy to merge PRs fixing fsspec handling, but in general I recommend using the non-fsspec API)

PR #664, opened 2026-04-11
feat!: open_reader accepts optional size to skip HEAD request

fvaleye:
# Summary
Add an optional `size` parameter to `obstore.open_reader` / `open_reader_async` that lets callers skip the `HEAD` request used to fetch the file size when the size is already known from external metadata.

[`create_reader`](https://github.com/developmentseed/obstore/blob/996978aeda7f298b99df441338fcfee685d2bcce/obstore/src/buffered.rs#L50-L60) currently HEADs the object solely to populate an `ObjectMeta`, but [`BufReader::with_capacity`](https://github.com/apache/arrow-rs-object-store/blob/62592027cb7d33ed843bc14fb77d7fc32f13ace5/src/buffered.rs#L94-L101) consumes only `meta.location` and `meta.size` from it. There is a network round-trip per file to fetch a single `u64`.

The same pattern already exists one layer up in the arrow-rs workspace: [`parquet::arrow::async_reader::ParquetObjectReader::with_file_size(u64)`](https://github.com/apache/arrow-rs/blob/68851ef953fd771cc310203c446e54145d4407e1/parquet/src/arrow/async_reader/store.rs#L88-L102) accepts an optional size hint as `Option<u64>`. Python-side similar examples exist in [`fsspec.AbstractBufferedFile(size=None)`](https://github.com/fsspec/filesystem_spec/blob/0bb819cf923b8680ac473f6a2faf492ab4d3602f/fsspec/spec.py#L1864) and [`pyarrow.dataset.FileFormat.make_fragment(file_size=None)`](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.FileFormat.html#pyarrow.dataset.FileFormat.make_fragment).

When `size` is provided, `create_reader` constructs the `ObjectMeta` directly with the hint and skips the `HEAD` entirely. When `size` is `None`, behavior is the same as before.

# Notes
The caller is responsible for the `size` value being accurate.
Failure modes:
- A hint larger than the actual file surfaces as an `OSError` at read time, when a range request past EOF is issued (`Invalid range`).
- A hint smaller than the actual file causes silent truncation: the reader treats the hint as authoritative EOF.

### DIFF
--- a/obstore/src/buffered.rs
+++ b/obstore/src/buffered.rs
@@ -2,11 +2,13 @@ use std::io::SeekFrom;
 use std::sync::Arc;
 
 use bytes::Bytes;
+use chrono::{DateTime, Utc};
+use indexmap::IndexMap;
 use object_store::buffered::{BufReader, BufWriter};
 use object_store::{ObjectMeta, ObjectStore, ObjectStoreExt};
-use pyo3::exceptions::{PyDeprecationWarning, PyIOError, PyStopAsyncIteration, PyStopIteration};
+use pyo3::exceptions::{PyIOError, PyStopAsyncIteration, PyStopIteration};
 use pyo3::prelude::*;
-use pyo3::types::PyString;
+use pyo3::types::{PyDict, PyString};
 use pyo3::{intern, IntoPyObjectExt};
 use pyo3_async_runtimes::tokio::{future_into_py, get_runtime};
 use pyo3_bytes::PyBytes;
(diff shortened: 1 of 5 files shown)


obstore/src/buffered.rs
```diff
@@ -92,8 +111,19 @@ impl PyReadableFile {
...
+        if !self.was_hinted {
+            dict.insert(
+                "last_modified",
+                self.meta.last_modified.into_bound_py_any(py)?,
+            );
+        }
+        dict.insert("size", self.meta.size.into_bound_py_any(py)?);
+        dict.insert("e_tag", self.meta.e_tag.clone().into_bound_py_any(py)?);
+        dict.insert("version", self.meta.version.clone().into_bound_py_any(py)?);
+        dict.into_pyobject(py)
```
kylebarron: I don't like this. This breaks the contract that `ObjectMeta` always returns a valid object:
https://github.com/developmentseed/obstore/blob/996978aeda7f298b99df441338fcfee685d2bcce/obstore/python/obstore/_buffered.pyi#L90-L92

I think it would be better to make this breaking and remove the `meta` attribute entirely. (Perhaps there's a way to express the deprecation for a release cycle?)

obstore/src/buffered.rs
```diff
@@ -15,62 +17,79 @@ use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncSeekExt, AsyncWriteExt, Line
...
+            .await
+            .map_err(PyObjectStoreError::ObjectStoreError)?,
+    };
     Ok((BufReader::with_capacity(store, &meta, capacity), meta))
 }
 
 #[pyclass(name = "ReadableFile", frozen)]
 pub(crate) struct PyReadableFile {
     reader: Arc<Mutex<BufReader>>,
     meta: ObjectMeta,
```
kylebarron: I think I'm realizing that `ObjectMeta` is the wrong abstraction for a `ReadableFile` to hold. 

It was easy to store that initially, when we always made the HEAD request ourselves, but it looks like `ReadableFile` doesn't actually use the `meta` in any way other than for the size. 

So I think that `PyReadableFile` should only store
```rs
struct PyReadableFile {
    reader: Arc<Mutex<BufReader>>,
    size: u64,
    r#async: bool
}
```
And then we don't have to store `was_hinted` either.

Users who want the other fields on `ObjectMeta` can make the `head` request themselves, persist the other fields, and then pass the `size` into `open_reader`

kylebarron (changes requested):
Thanks for the PR!

I think best to no longer store `meta: ObjectMeta` internally.

Curious if it's worth making the breaking changes non-breaking and deprecated for a release cycle, or if it's worth just removing `.meta` now. What do you think?

obstore/src/buffered.rs
fvaleye: Agreed. I only kept `ObjectMeta` to preserve the existing `.meta` attribute and avoid a breaking change.
Happy to drop it from `PyReadableFile` and store just reader/size/async as you suggested, which also lets us get rid of `was_hinted.`

For the transition, we could deprecate `.meta` for one release cycle rather than removing it right now.
We can mark it with `@deprecated` (via `typing_extensions.deprecated`). Users will receive a warning, and it will be removed in the next release.

Does that sound good to you?

kylebarron: Yes that sounds great.

How about:

- First make a PR to just add a deprecation warning and `@deprecated` onto the `.meta` attribute. We can release this in a patch release?
- Then make a PR (this PR) to remove `.meta` and store only size. Then we can publish that in a new minor (breaking) release. Ideally I want to get https://github.com/developmentseed/obstore/pull/545 finally merged in the next release too

kylebarron (approved):
Thank you!

kylebarron:
Forgot to mark this as breaking because it removes the `meta` param

PR #672, opened 2026-04-22
fix: Fix path parsing for copy, rename

### LINKED ISSUE #671: obs.copy (and obs.rename) fails on non-ASCII filenames (e.g. with accented characters)
obs.copy() percent-encodes the source path internally ("café.txt" → "caf%C3%A9.txt", as we see in the traceback). Consequently it raises FileNotFoundError because the file is actually stored as "café.txt".
obs.rename() does the same, causing the same error. The other methods such as obs.head or obs.list work.

Here is a minimal reproduction for obs.copy that you can run:

```python
import tempfile

import obstore as obs
from obstore.store import LocalStore

with tempfile.TemporaryDirectory() as tmpdir:
    store = LocalStore(tmpdir)

    store.put("café.txt", b"hello")

    # Sanity check: head() finds the file just fine.
    assert obs.head(store, "café.txt")["path"] == "café.txt"

    # Bug: copy() percent-encodes the source path and raises FileNotFoundError.
    obs.copy(store, "café.txt", "new.txt")
```

The error is the following:

```
FileNotFoundError: Object at location /tmp/tmpnsnnqrot/caf%C3%A9.txt not found: No such file or directory (os error 2)

Debug source:
NotFound {
    path: "/tmp/tmpnsnnqrot/caf%C3%A9.txt",
    source: Os {
        code: 2,
        kind: NotFound,
        message: "No such file or directory",
    },
}
```

kylebarron:
### Change list

- Use `PyPath` instead of `String` as parameter input for `copy`, `list`, and `rename`
- Add tests
- Bump pyo3 version from 0.28.2 to 0.28.3 to fix CI

Closes https://github.com/developmentseed/obstore/issues/671

The primary change to avoid auto-percent-encoding of paths was done in https://github.com/developmentseed/obstore/pull/524, but it looks like we missed path inputs for copy/rename.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+*.dylib
+*.dylib.yml
+Info.plist
 *.so
 *.whl
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1520,9 +1520,9 @@ dependencies = [
 
 [[package]]
 name = "pyo3"
-version = "0.28.2"
+version = "0.28.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf85e27e86080aafd5a22eae58a162e133a589551542b3e5cee4beb27e54f8e1"
+checksum = "91fd8e38a3b50ed1167fb981cd6fd60147e091784c427b8f7183a7ee32c31c12"
 dependencies = [
  "chrono",
  "chrono-tz",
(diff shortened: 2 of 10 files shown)


kylebarron:
Since the main breaking change was in #524 , I think this can be considered a bugfix and go in a patch release.

PR #724, opened 2026-06-16
feat: Expose proxy and custom-CA client options

### LINKED ISSUE #428: Support custom root certificates
Using a similar approach as in #427  and #426 , we could expose [`with_custom_root_certificate`](https://docs.rs/object_store/latest/object_store/client/struct.ClientOptions.html#method.with_root_certificate)

louisnow:
Fixes #428 

My use case is that there's an internal corp proxy which has a custom certificate which then connects to an s3 compatible API service.

### DIFF
--- a/pyo3-object_store/src/client.rs
+++ b/pyo3-object_store/src/client.rs
@@ -10,6 +10,7 @@ use pyo3::types::{PyBytes, PyDict, PyString};
 
 use crate::config::PyConfigValue;
 use crate::error::PyObjectStoreError;
+use crate::PyObjectStoreResult;
 
 /// A wrapper around one or more `Certificate`s parsed from PEM input.
 ///
@@ -19,6 +20,13 @@ use crate::error::PyObjectStoreError;
 struct PyCertificate {
     pem: Vec<u8>,
     certificates: Vec<Certificate>,
+}
+
+impl PyCertificate {
+    fn new(pem: Vec<u8>) -> PyObjectStoreResult<Self> {
+        let certificates = Certificate::from_pem_bundle(&pem)?;
+        Ok(Self { pem, certificates })
+    }
 }
 
 impl PartialEq for PyCertificate {
@@ -36,9 +44,7 @@ impl<'py> FromPyObject<'_, 'py> for PyCertificate {
         } else {
             obj.extract::<String>()?.into_bytes()
         };
-        let certificates =
-            Certificate::from_pem_bundle(&pem).map_err(PyObjectStoreError::ObjectStoreError)?;
-        Ok(Self { pem, certificates })
+        Ok(Self::new(pem)?)
     }
 }
 
--- a/pyo3-object_store/src/client.rs
+++ b/pyo3-object_store/src/client.rs
@@ -11,12 +11,44 @@ use pyo3::types::{PyBytes, PyDict, PyString};
 use crate::config::PyConfigValue;
 use crate::error::PyObjectStoreError;
 
-fn extract_pem(value: &Bound<'_, PyAny>) -> PyResult<Vec<u8>> {
-    if let Ok(bytes) = value.extract::<PyBackedBytes>() {
-        Ok(bytes.as_ref().to_vec())
-    } else {
-        let s = value.extract::<PyBackedStr>()?;
-        Ok(s.as_bytes().to_vec())
+/// A wrapper around one or more `Certificate`s parsed from PEM input.
+///
+/// The original PEM is retained so the value round-trips through
+/// [`IntoPyObject`]; parsing happens once, on extraction.
+#[derive(Clone, Debug)]
+struct PyCertificate {
+    pem: Vec<u8>,
+    certificates: Vec<Certificate>,
+}
+
+impl PartialEq for PyCertificate {
+    fn eq(&self, other: &Self) -> bool {
+        self.pem == other.pem
+    }
+}
+
+impl<'py> FromPyObject<'_, 'py> for PyCertificate {
+    type Error = PyErr;
+
+    fn extract(obj: Borrowed<'_, 'py, pyo3::PyAny>) -> PyResult<Self> {
+        let pem = if let Ok(bytes) = obj.extract::<PyBackedBytes>() {
+            bytes.as_ref().to_vec()
+        } else {
+            obj.extract::<PyBackedStr>()?.as_bytes().to_vec()
+        };
+        let certificates =
+            Certificate::from_pem_bundle(&pem).map_err(PyObjectStoreError::ObjectStoreError)?;
+        Ok(Self { pem, certificates })
+    }
+}
+
+impl<'py> IntoPyObject<'py> for &PyCertificate {
+    type Target = PyBytes;
+    type Output = Bound<'py, PyBytes>;
+    type Error = PyErr;
+
+    fn into_pyobject(self, py: Python<'py>) -> Result<Self::Output, Self::Error> {
+        Ok(PyBytes::new(py, &self.pem))
     }
 }
 
@@ -59,9 +91,7 @@ impl<'py> IntoPyObject<'py> for &PyClientConfigKey {
 pub struct PyClientOptions {
     string_options: HashMap<PyClientConfigKey, PyConfigValue>,
     default_headers: Option<PyHeaderMap>,
-    // Stored as raw PEM so it round-trips through `IntoPyObject`; parsed on the
-    // way into `ClientOptions`.
-    root_certificate: Option<Vec<u8>>,
+    root_certificate: Option<PyCertificate>,
 }
 
 impl<'py> FromPyObject<'_, 'py> for PyClientOptions {
@@ -81,7 +111,9 @@ impl<'py> FromPyObject<'_, 'py> for PyClientOptions {
                 let key = key.extract::<PyBackedStr>()?;
                 match &*key {
                     "default_headers" => default_headers = Some(value.extract::<PyHeaderMap>()?),
-                    "root_certificate" => root_certificate = Some(extract_pem(&value)?),
+                    "root_certificate" => {
+                        root_certificate = Some(value.extract::<PyCertificate>()?)
+                    }
                     _ => return Err(PyValueError::new_err(format!("Invalid key: {key}."))),
                 }
             }
@@ -105,8 +137,8 @@ impl<'py> IntoPyObject<'py> for PyClientOptions {
         if let Some(headers) = self.default_headers {
             dict.set_item("default_headers", headers)?;
         }
-        if let Some(pem) = self.root_certificate {
-            dict.set_item("root_certificate", PyBytes::new(py, &pem))?;
+        if let Some(certificate) = &self.root_certificate {
+            dict.set_item("root_certificate", certificate)?;
         }
         Ok(dict)
     }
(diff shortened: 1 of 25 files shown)


pyo3-object_store/src/client.rs
```diff
@@ -2,15 +2,24 @@ use std::collections::HashMap;
...
 use crate::error::PyObjectStoreError;
 
+fn extract_pem(value: &Bound<'_, PyAny>) -> PyResult<Vec<u8>> {
+    if let Ok(bytes) = value.extract::<PyBackedBytes>() {
+        Ok(bytes.as_ref().to_vec())
+    } else {
+        let s = value.extract::<PyBackedStr>()?;
+        Ok(s.as_bytes().to_vec())
+    }
+}
```
kylebarron: If we accept PyBackedStr only to convert it to `Vec<u8>`, I think it would be cleaner to only accept binary input.

kylebarron: Actually I just refreshed myself and a PEM is an ASCII string, is that right? So maybe allowing str input would be useful.

I see that `object-store` supports [`DER` input](https://docs.rs/object_store/latest/object_store/client/struct.Certificate.html#method.from_der) to create a Certificate. Would that ever be useful? Should we have a new top-level `Certificate` class in Python that users can construct from either a PEM or a DER?

I don't know how relatively popular is each (I've only ever seen PEM in the wild)

kylebarron: Regardless, I think it would be good to make a `PyCertificate` wrapper struct around `Certificate`, and then we can implement `FromPyObject` and `IntoPyObject` on that.

pyo3-object_store/src/client.rs
```diff
@@ -50,6 +59,9 @@ impl<'py> IntoPyObject<'py> for &PyClientConfigKey {
 pub struct PyClientOptions {
     string_options: HashMap<PyClientConfigKey, PyConfigValue>,
     default_headers: Option<PyHeaderMap>,
+    // Stored as raw PEM so it round-trips through `IntoPyObject`; parsed on the
+    // way into `ClientOptions`.
+    root_certificate: Option<Vec<u8>>,
```
kylebarron: As mentioned above, this would switch to `Option<PyCertificate>` which then would be readable enough that it wouldn't require a comment.

pyo3-object_store/src/client.rs
```diff
@@ -91,6 +105,9 @@ impl<'py> IntoPyObject<'py> for PyClientOptions {
         if let Some(headers) = self.default_headers {
             dict.set_item("default_headers", headers)?;
         }
+        if let Some(pem) = self.root_certificate {
+            dict.set_item("root_certificate", PyBytes::new(py, &pem))?;
+        }
```
kylebarron: Nit: It'll be slightly cleaner if we define `IntoPyObject` on `PyCertificate`, because then the Certificate -> Python serialization is defined only in one place.

pyo3-object_store/src/client.rs
```diff
@@ -120,7 +142,13 @@ impl From<PyClientOptions> for ClientOptions {
             options = options.with_default_headers(headers.0);
         }
 
-        options
+        if let Some(pem) = value.root_certificate {
+            for certificate in Certificate::from_pem_bundle(&pem)? {
```
kylebarron: If we define a `PyCertificate` struct, then the conversion from Python object to Certificate is done at import time, and that means this PyClientOptions -> ClientOptions can stay as infallible.

tests/store/test_client_options.py
```diff
@@ -0,0 +1,129 @@
...
+LwIDAQABo1MwUTAdBgNVHQ4EFgQUJGhd1ORfcw1Ov3/fljWfEoUJSx8wHwYDVR0j
+BBgwFoAUJGhd1ORfcw1Ov3/fljWfEoUJSx8wDwYDVR0TAQH/BAUwAwEB/zANBgkq
+hkiG9w0BAQsFAAOCAQEAurAH4Gvomh4QffpZf6s3/TPXHGM2wuuuOKB7QD7sfp8V
+nlFthbcpEd0SRaH2T4fIeVDuHcMx/F7GYBaVOjXqRq9N6+zeutpPvu7YAoE4zbLz
+Wnrn2fZG5uAO+HW26QXOsZU2zJHpHzZWZ3G1dV/C97k8hmSjkH7OOZSlZ/qIN4+p
+5t5qBUJek66luxnEyfOdFitiJe9Ri6sj2ffT0aXejZCu2boO+3Szm6boLu3kCC7l
+Latj8uDxEm6HsD4gxn61zMmmSjitTJYJt+lW8+3tlSf17tXd1TLG/cF/0gD28OhL
+UmOHI1HSNtVKwotBkpOrlby1hWOX7IMtqPRmEJzz6Q==
+-----END CERTIFICATE-----
+"""
```
kylebarron, line 29: Just for my knowledge, how were these test cases defined? Is this random input?

tests/store/test_client_options.py
```diff
@@ -0,0 +1,129 @@
...
+hkiG9w0BAQsFAAOCAQEAiJvYmLrqG/bMWyb3u77tLSfwxGftIo5HApgKt6vgtUZr
+0qhgf1scto1lIworYX+MJe/dqHVZC6IjiXOO77HCJlOQoPd3byvV+fu3sU1P79ah
+tIFKyRN6cGfAXlPYhR3oB30KEHIMZKve/Y9c1IGv7cogO0/VDXXTe3YcWDLtjzd8
+2wGP3vlKhbzV2ZigRtO37fCTBGJRlpmUT+KWQ5roSSbutyYNhAAAOMoM+M6uD4Mm
+H3ZwuubjcsNBfC/YQabjFuDrIUAYUS7CWDVv1Qi/26PlRQJEFoVayDT4aO86cAop
+Tg1+pQESP0OOub4TswJ0qs0Xx2umJdos/XOaf3xfpQ==
+-----END CERTIFICATE-----
+"""
+
+CERT_BUNDLE = CERT_1 + CERT_2
```
kylebarron, line 52: https://docs.rs/object_store/latest/object_store/client/struct.Certificate.html#method.from_pem_bundle

Is this concept similar to the PEM bundle linked?

kylebarron (commented):
Thanks for the PR! This is a good start

tests/store/test_client_options.py
louisnow: I generated them using openssl, I've added the command as a comment now
```
openssl req -x509 -newkey rsa:2048 -nodes -days 36500 -subj /CN=obstore-test-ca-N
```

tests/store/test_client_options.py
louisnow: Yeah, concatenation looks fine to me

```
One advantage with text-based encodings is that they are easy to
modify using common text editors; for example, a user may concatenate
several certificates to form a certificate chain with copy-and-paste
operations. The tradition within the RFC series can be traced back to 
Privacy-Enhanced Mail (PEM)
```

https://datatracker.ietf.org/doc/html/rfc7468#section-1

```
... a certificate which has the same subject and issuer, sometimes called a root certificate. The certificates should just be concatenated together in the certificate file. 
```

https://docs.python.org/3/library/ssl.html#certificate-chains

pyo3-object_store/src/client.rs
louisnow: Thanks, I've made a PyCertificate wrapper struct now

louisnow: > Should we have a new top-level Certificate class in Python that users can construct from either a PEM or a DER?

https://docs.python.org/3/library/ssl.html#ssl-certificates

```
Python uses files to contain certificates. They should be formatted as “PEM” (see [RFC 1422](https://datatracker.ietf.org/doc/html/rfc1422.html)), which is a base-64 encoded form wrapped with a header line and a footer line:
```

PEM is quite common, I'm not sure about needing to expose a Certificate class just for this library, let the author pass in str/bytes which is easily obtainable by just reading the certificate file from disk.

tests/store/test_client_options.py
kylebarron: Cool, thanks. I've never touched custom certificates myself

pyo3-object_store/src/client.rs
```diff
@@ -2,15 +2,56 @@ use std::collections::HashMap;
...
-use pyo3::types::{PyDict, PyString};
+use pyo3::types::{PyBytes, PyDict, PyString};
 
 use crate::config::PyConfigValue;
 use crate::error::PyObjectStoreError;
 
+/// A wrapper around one or more `Certificate`s parsed from PEM input.
+///
+/// The original PEM is retained so the value round-trips through
+/// [`IntoPyObject`]; parsing happens once, on extraction.
```
kylebarron, line 18: 👍  I see, it looks like there's no method on `Certificate` to go back to a `PEM`.

pyo3-object_store/src/client.rs
```diff
@@ -2,15 +2,56 @@ use std::collections::HashMap;
...
+        self.pem == other.pem
+    }
+}
+
+impl<'py> FromPyObject<'_, 'py> for PyCertificate {
+    type Error = PyErr;
+
+    fn extract(obj: Borrowed<'_, 'py, pyo3::PyAny>) -> PyResult<Self> {
+        let pem = if let Ok(bytes) = obj.extract::<PyBackedBytes>() {
+            bytes.as_ref().to_vec()
```
kylebarron: If we're immediately going to convert to a `vec`, I don't think there's any usefulness of going through `PyBackedBytes`. You can just extract to `Vec<u8>` directly.

pyo3-object_store/src/client.rs
```diff
@@ -2,15 +2,56 @@ use std::collections::HashMap;
...
+        } else {
+            obj.extract::<PyBackedStr>()?.as_bytes().to_vec()
```
kylebarron: And similarly I think you can extract as a `String`, and then I assume `String -> Vec<u8>` is zero-copy?

pyo3-object_store/src/client.rs
```diff
@@ -2,15 +2,56 @@ use std::collections::HashMap;
...
 
+/// A wrapper around one or more `Certificate`s parsed from PEM input.
+///
+/// The original PEM is retained so the value round-trips through
+/// [`IntoPyObject`]; parsing happens once, on extraction.
+#[derive(Clone, Debug)]
+struct PyCertificate {
+    pem: Vec<u8>,
+    certificates: Vec<Certificate>,
+}
```
kylebarron, line 23: One more style nit here: since the pem is the only part actually used to construct `PyCertificate`, could we add `PyCertificate::new` which takes in just `pem: Vec<u8>`? Then the certificate creation would happen inside this `::new`, which makes it even clearer that the _only_ thing happening in the extraction from Python is taking in a string or bytes, and passing it to `PyCertificate::new`

tests/store/test_client_options.py
kylebarron: I think I'm asking because I want to know: would we ever want to expose to the Python user the difference between 	`Certificate::from_pem` and `Certificate::from_pem_bundle`? 

In the implementation here, we hard-code `Certificate::from_pem` in the extraction in `FromPyObject`, and then we couldn't add `Certificate::from_pem_bundle` in the future. 

If we wanted to support both, I think we'd need to expose a `Certificate` class to python with both constructors.

But I don't know enough about the difference between the two constructors to know whether an end user might ever want both

kylebarron (commented):
Thanks! Just a few more comments

PR #745, opened 2026-07-03
fix: strip prefix from `get_opts` result in `MaybePrefixedStore`

### LINKED ISSUE #744: `open_reader` double-applies a store `prefix=`
On a store created with a `prefix=` (e.g. `S3Store(prefix="P")`), `obstore.open_reader(store, "key")` resolves the object at `P/P/key` instead of `P/key`, so every buffered read of a prefixed store fails with a 404. `get`, `put`, and `list` on the same store are correct (single prefix). Regressed in 0.9.0; still present in 0.11.0.

## Environment

- obstore 0.11.0 also reproduced on 0.9.0.
- Last good: 0.8.2

## Reproduction

Standalone `uv run repro.py` (deps declared inline; moto provides a real S3 backend). The object is written once at `P/key`; swap the `obstore` pin to `0.8.2` to see the correct behavior.

```python
# /// script
# requires-python = ">=3.11"
# dependencies = ["obstore==0.11.0", "boto3==1.43.22", "moto[server]==5.2.1"]
# ///
import boto3, obstore
from moto.server import ThreadedMotoServer

server = ThreadedMotoServer(port=0); server.start()
host, port = server.get_host_and_port()
endpoint = f"http://{host}:{port}"

s3 = boto3.client("s3", endpoint_url=endpoint, region_name="us-east-1",
                  aws_access_key_id="x", aws_secret_access_key="x")
s3.create_bucket(Bucket="bucket")
s3.put_object(Bucket="bucket", Key="P/key", Body=b"hello")  # object lives at P/key

store = S3Store(bucket="bucket", prefix="P", endpoint=endpoint, region="us-east-1",
                access_key_id="x", secret_access_key="x",
                client_options={"allow_http": True}, virtual_hosted_style_request=False)

print("list :", [o["path"] for b in obstore.list(store) for o in b])  # ['key']
print("head :", obstore.head(store, "key")["path"])                    # 'P/key'  <- should be 'key'
print("get  :", bytes(obstore.get(store, "key").bytes()))              # b'hello'
try:
    print("read :", obstore.open_reader(store, "key").read())          # -> GET P/P/key -> 404
except Exception as e:
    print("read :", str(e).splitlines()[0])
server.stop()
```

## Output

**obstore 0.11.0 (bug)** — moto also logs `HEAD /bucket/P/key` then `GET /bucket/P/P/key`:

```
list : ['key']       # prefix stripped (correct)
head : P/key         # <- prefix NOT stripped (root cause)
get  : b'hello'      # correct
read : Object at location P/P/key not found: ... GET .../bucket/P/P/key ... 404
```

**obstore 0.8.2 (correct)** — GET goes to `/bucket/P/key`:

```
list : ['key']
head : key
get  : b'hello'
read : Bytes(b'hello')
```

## Expected

`open_reader(store, "key")` on a store with `prefix="P"` should read `P/key` (as `get`/`put`/`list` do), and `head(...)["path"]` should return `"key"` (prefix-relative), consistent with `list()`.

blinkybool:
Fixes #744.

`MaybePrefixedStore::get_opts` returned an `ObjectMeta` whose `location` still carried the store prefix, unlike `list`/`list_with_delimiter` (which `strip_meta`). Since `head` delegates to `get_opts`, and the buffered reader (`open_reader`) issues its follow-up range read against the returned `meta.location`, reads on a store with a `prefix=` resolved at `prefix/prefix/key` and 404'd.

This mirrors object_store's own fix for the same issue in its `PrefixStore` ([apache/arrow-rs-object-store#686](https://github.com/apache/arrow-rs-object-store/pull/686)) — the copy vendored here predates it.

Kept to the one-line fix to match the crate's current lack of tests; #744 has a standalone `uv run` repro. Happy to add a regression test if you'd like one.

---

🤖 Investigated and drafted with [Claude Code](https://claude.com/claude-code); reviewed by me.

### DIFF
--- a/pyo3-object_store/src/prefix.rs
+++ b/pyo3-object_store/src/prefix.rs
@@ -148,7 +148,9 @@ impl<T: ObjectStore> ObjectStore for MaybePrefixedStore<T> {
 
     async fn get_opts(&self, location: &Path, options: GetOptions) -> Result<GetResult> {
         let full_path = self.full_path(location);
-        self.inner.get_opts(&full_path, options).await
+        let mut result = self.inner.get_opts(&full_path, options).await?;
+        result.meta = self.strip_meta(result.meta);
+        Ok(result)
     }
 
     async fn get_ranges(&self, location: &Path, ranges: &[Range<u64>]) -> Result<Vec<Bytes>> {


kylebarron:
Could you add a test for this please?

PR #757, opened 2026-07-23
feat: Support R2 URLs with jurisdictions

Kharacternyk:
Relevant upstream docs: https://developers.cloudflare.com/r2/reference/data-location/#jurisdictional-restrictions

### DIFF
--- a/docs/examples/r2.md
+++ b/docs/examples/r2.md
@@ -22,7 +22,7 @@ It's easy to read and write data to and from Cloudflare R2 with obstore's [`S3St
     access_key_id = "..."
     secret_access_key = "..."
     store = S3Store.from_url(
-        "https://f0b62eebfbdde1133378bfe3958325f6.r2.cloudflarestorage.com/ kylebarron-public",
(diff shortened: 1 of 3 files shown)


Kharacternyk:
I don't understand from the GitHub action logs which of my changes are triggering the pre-commit failure

kylebarron:
Thanks for the PR.

In regards to the pre-commit error, I think the issue is probably that it's not pinned to a specific version of ruff, and that a new version of ruff added that lint.

Separately, the code you're touching in this PR, `parse_url` is vendored from upstream https://github.com/apache/arrow-rs-object-store. For maintenance reasons, I don't want our copy to diverge from upstream. 

So if you could make a PR to upstream `object-store`, that would be great, and once that is merged, then we can merge this here as well.

Kharacternyk:
Hi @kylebarron, this has been merged upstream: https://github.com/apache/arrow-rs-object-store/pull/815

kylebarron:
great thanks, if you could merge in main, that should fix the CI

Kharacternyk:
Yep, all green now

kylebarron (approved):
Thanks

PR #760, opened 2026-08-03
fix: expose s3-specific endpoint config

### LINKED ISSUE #753: `AWS_ENDPOINT_URL_S3` silently overrides the explicit `endpoint` kwarg
**obstore 0.11.0, Python 3.13, Linux**

When `AWS_ENDPOINT_URL_S3` is set in the environment, `S3Store(..., endpoint=...)` ignores the kwarg and uses the env var. boto3 does the opposite: an explicit `endpoint_url` beats the env var. An explicitly passed argument losing to ambient environment is surprising, and there's no warning that it happened.

## Repro

```python
import os
os.environ["AWS_ENDPOINT_URL_S3"] = "https://env-endpoint.example"

import obstore

store = S3Store(
    "demo-bucket",
    region="us-east-1",
    endpoint="https://demo-bucket.kwarg-endpoint.example",
    virtual_hosted_style_request=True,
)
obstore.list(store).collect()
```

The list request goes to `http://env-endpoint.example/?list-type=2` — the kwarg
endpoint never appears.

## Expected

Explicit config wins over environment, matching boto3 and most config systems:

|                                     | boto3        | obstore 0.11        |
| ----------------------------------- | ------------ | ------------------- |
| `endpoint` kwarg vs `AWS_ENDPOINT_URL_S3` | kwarg wins   | env wins            |
| precedence documented               | yes          | not that I found    |

If env-wins is intentional, a documented note on `S3Store` plus a warning when
both are set and disagree would be helpful.

Guflly:
Fixes #753.

This adds an s3_endpoint option so callers can explicitly override AWS_ENDPOINT_URL_S3. The alias round-trips through config and pickling, and the regression covers the environment precedence case.

Draft #755 touches the same config-key conversion code, so this may need a small rebase if it lands first.

Tests:
- cargo test --all --all-features
- cargo clippy on pyo3-object_store
- Ruff and Pyright
- endpoint regression against a local HTTP server

### DIFF
--- a/pyo3-object_store/src/aws/store.rs
+++ b/pyo3-object_store/src/aws/store.rs
@@ -230,11 +230,7 @@ impl<'py> FromPyObject<'_, 'py> for PyAmazonS3ConfigKey {
 
     fn extract(obj: Borrowed<'_, 'py, pyo3::PyAny>) -> PyResult<Self> {
         let s = obj.extract::<PyBackedStr>()?.to_lowercase();
-        let key = if s == "s3_endpoint" {
-            AmazonS3ConfigKey::S3Endpoint
-        } else {
-            s.parse().map_err(PyObjectStoreError::ObjectStoreError)?
-        };
+        let key = s.parse().map_err(PyObjectStoreError::ObjectStoreError)?;
         Ok(Self(key))
     }
 }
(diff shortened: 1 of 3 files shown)


pyo3-object_store/src/aws/store.rs
```diff
@@ -230,7 +230,11 @@ impl<'py> FromPyObject<'_, 'py> for PyAmazonS3ConfigKey {
 
     fn extract(obj: Borrowed<'_, 'py, pyo3::PyAny>) -> PyResult<Self> {
         let s = obj.extract::<PyBackedStr>()?.to_lowercase();
-        let key = s.parse().map_err(PyObjectStoreError::ObjectStoreError)?;
+        let key = if s == "s3_endpoint" {
+            AmazonS3ConfigKey::S3Endpoint
+        } else {
+            s.parse().map_err(PyObjectStoreError::ObjectStoreError)?
+        };
```
kylebarron: I don't want to make special cases here.

We can document that the allowed key is `aws_endpoint_url_s3`, or we can expand the allowed upstream cases here:
https://github.com/apache/arrow-rs-object-store/blob/84d24eb8efcec9448566de09e94d2d4b74b21ebe/src/aws/builder.rs#L517

Guflly: Switched this to the existing `aws_endpoint_url_s3` key and removed the local special case. The PR now only updates the S3 config type and docs.

kylebarron (approved):
Thanks

PR #775, opened 2026-09-04
ci: add Windows ARM64 wheels

ndabas:
As the title says -- the official CPython releases for Windows ARM64 start from 3.11 so we build the abi3 wheel and a 3.14t one. I've adjusted the matrix a bit so we don't need to put in conditionals for skipping 3.10, which should still be built for Windows x64.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -130,6 +130,10 @@ jobs:
         platform:
           - runner: windows-latest
             target: x64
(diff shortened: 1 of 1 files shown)


kylebarron:
Thanks! Were you able to create a wheel build CI run on your branch to test this?

ndabas:
@kylebarron Yes, indeed I was: https://github.com/ndabas/obstore/actions/runs/33865637930

kylebarron:
Thanks

